PR #4, opened 2015-01-27
Windows compatibility fixes

insin:
* child_process.spawn ignores PATHEXT on Windows, see joyent/node#2318
* find doesn't normalise path separators, added window variants to file filter RegExp
* `npm test` doesn't work on Windows with bare path to .js file


feross:
This is amazing work! Thanks!

feross:
Released as 1.1.0.

PR #14, opened 2015-01-30
add --rcpath option

Flet:
This PR adds a new option --rcpath (or -r for short). This allows the user to provide an alternative directory to use which could hold one or many .rc files.

example:
```bash
standard --rcpath my/custom/rules
```

- Additional logic has been added to ensure `standard` will fallback to the standard .rc files in its own rc directory if they don't exist. This allows the user to just override a single .rc file if they wish.
- An additional test function has been added which will use a custom .jshintrc file with "use strict" turned on. `standard --rcpath altrc` is then run againts the same set of git repos as before, and it "passes" only if errors come back.
- `cmd.js` has been switched to use `minimist` to handle multiple args nicely.

I've sensed a bit of a philosophical struggle on adding flags to turn on/off things for this module, and I agree that it should be opinionated. Adding this option could be a way for users to do their own thing or deviate without further complicating things. It also gives you a canned response for bikeshed issues: "use the --rcpath option if you don't like rule X."

The ulterier motive here is that this gives me a graceful way of handling Flet/semistandard#2 without it needed to be a full forked copy of this repo.

### DIFF
--- a/bin/altrc/.jshintrc
+++ b/bin/altrc/.jshintrc
@@ -1,0 +1,29 @@
+{
+  "asi": true,
+  "browser": true,
+  "devel": true,
+  "eqeqeq": true,
+  "eqnull": true,
+  "esnext": true,
+  "expr": true,
+  "globals": {
+    "chrome": false,
+    "FileList": false
+  },
+  "globalstrict": true,
+  "immed": true,
+  "latedef": "nofunc",
+  "laxbreak": true,
+  "loopfunc": true,
+  "newcap": true,
+  "noarg": true,
+  "node": true,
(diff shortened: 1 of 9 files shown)


Flet:
Ah shoot, you've been busy... I'll have to sync ths up :)

mafintosh:
We shouldn't support `rc` files in my opinion. Standard should be opinionated and unconfigurable. Otherwise we'll just end up moving all the bikeshedding from this repo to all the repos using standard.

dcousens:
No options

Flet:
Alrighty then. Shutting this PR down.

PR #19, opened 2015-02-04
removed duplicated rules from .jscsrc

Flet:
Removed 5 duplicated rules from .jscrc (pattern of previous commits seemed to favor eslint).

mixed spaces and tabs

multiline strings

unary after

unary before

cap constructor


feross:
LGTM!

feross:
Released as 2.4.1.

PR #23, opened 2015-02-04
Ignore coverage directory by default

### LINKED ISSUE #22: Exclude `coverage` directory
When I use `istanbul` in a project, it generates some js files in the `coverage` directory. This directory is in the Github-default .gitignore, and standard should probably ignore it by default too.

brandonhorst:
Fixes #22


feross:
Released as 2.4.2.

PR #27, opened 2015-02-05
change clone to --depth 1 in test.js

Flet:
Changing to `--depth 1` saves about 7MB of total downloading and and shaves about 5 seconds from testing time for me (downloading at 2.00 MiB/s),


feross:
Nice!

PR #30, opened 2015-02-06
Don't rely on node_modules/.bin path

feross:
It's unsafe to assume binaries will be installed into `node_modules/.bin`.

`npm dedupe` and npm@3 will try to move them to a higher-level `node_modules` when possible to simplify the dependency tree. [see the [npm blog](http://blog.npmjs.org/post/109921886740/a-belated-npm-weekly-3)]

Instead, let's find the location at runtime using the `PATH` that npm sets for us.

@othiym23 / @ljharb - This look right to you?
@insin - Could you test this out on Windows? Then I'll do a release.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,20 +7,13 @@ var path = require('path')
 var split = require('split')
 var uniq = require('uniq')
 
-var JSCS = path.join(__dirname, 'node_modules', '.bin', 'jscs')
 var JSCS_RC = path.join(__dirname, 'rc', '.jscsrc')
 var JSCS_REPORTER = path.join(__dirname, 'lib', 'jscs-reporter.js')
 var JSCS_REPORTER_VERBOSE = path.join(__dirname, 'lib', 'jscs-reporter-verbose.js')
 
-var ESLINT = path.join(__dirname, 'node_modules', '.bin', 'eslint')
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -123,12 +116,30 @@ module.exports = function standard (opts) {
...
+      eslint: findBinPath.bind(undefined, 'eslint'),
+      jscs: findBinPath.bind(undefined, 'jscs')
+    }, cb)
+  }
+
+  function findBinPath (bin, cb) {
+    var opts = { cwd: __dirname }
+    cp.exec('npm run --silent which-' + bin, opts, function (err, stdout, stderr) {
+      if (err) return cb(err)
+      cb(null, stdout.toString().replace(/\n/g, ''))
```
ljharb: I prefer `String(stdout)` since that guarantees the return type, whereas `toString` does not - but since it's from `child_process.exec` it doesn't matter that much.

ljharb:
LGTM 👍

othiym23:
LGTM. We should figure out how to make this process less painful, though.

maxogden:
Is this relevant? https://www.npmjs.com/package/npm-execspawn

feross:
@maxogden Hmm, yeah, that looks like it would approximate `npm run`'s PATH.

feross:
Btw, this change is released as 2.4.5.

PR #35, opened 2015-02-14
always ignore node_modules

alexanderGugel:
Since it's not that uncommon to have e.g. an examples directory (see [react](https://github.com/facebook/react/tree/master/examples/basic-commonjs)) with a `package.json`. Therefore I think it's a good idea to *always* ignore `node_modules`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,7 +16,7 @@ var ESLINT_REPORTER = path.join(__dirname, 'lib', 'eslint-reporter.js')
 var ESLINT_REPORTER_VERBOSE = path.join(__dirname, 'lib', 'eslint-reporter-verbose.js')
 
 var DEFAULT_IGNORE = [
-  'node_modules/**',
+  '**/node_modules/**',
   '.git/**',
   '**/*.min.js',
   '**/bundle.js',


dcousens:
I see no issue with this, ACK

Flet:
This would not play nice with this method of organizing projects:
[Working without frameworks (Part 1)](https://medium.com/node-js-javascript/working-without-frameworks-part-1-b948f281f782)

>
__The Under-appreciated node_modules directory__
Node’s require algorithm is applied to any directory called node_modules, not just the one at the root of your project! Lately, I’ve been putting my application specific code into a folder called app/node_modules.

For this reason I would advocate for only ignoring `node_modules` at the root of the package.

yoshuawuyts:
When using [`linklocal`](https://github.com/timoxley/linklocal) and npm's local modules (`file:`) it's nice if all instances of `node_modules` are ignored. Maybe having a flag that can be toggled to un-ignore `node_modules` would be the correct solution? That way both methods of managing application dependencies can be supported.

Flet:
Its also possible to achive this without a code change currently by adding this to a modules `package.json`

```
 "standard": {
    "ignore": ["**/node_modules/**"]
  }
```

There is not currently a way to "include" a path that is currently excluded.

feross:
I'm inclined to recommend @Flet's solution for the time being, since it's not possible to un-ignore a directory at the moment. But I'll leave this PR open for a while in case others have opinions.

/cc @mafintosh @maxogden

mafintosh:
I have no strong opinions about this. If possible we could ignore node_modules if a package.json is present in the same directory

feross:
cool, i'll accept a PR that implements @mafintosh's suggestion.

Flet:
Even with @mafintosh's suggestion it will not suffice. Quoting the article above (talking about items in an apps/node_modules directory):
>
Each part of a program that isn’t published, I treat as a program module. I’ll create a README.md file for it to explain its thesis and usage. I create a package.json for it (mainly for use by browserify). It will have an index.js and maybe even some sub-directories. These modules then become require-able without hard coding their path (this also makes refactors easier).

Is this application layout is too obscure/strange/anti-pattern to consider legit? 

Are there more important patterns or use cases that warrant the borader ignore behavior?

Not fighting for it -- just curious.

alexanderGugel:
@Flet I think this is rather obscure and I don't really think this simplifies things:

Whenever I see something like `require('cool-module')` and I don't know what *cool-module* is about, I look it up on the npm registry (https://www.npmjs.com/package/whatever). I don't think this is a very uncommon workflow (yes, you could argue that I could have a look into the README in node_modules directly, but that way I wouldn't see the GitHub issues, popularity, if it's being maintained etc.). I can imagine that you might also run into some weird naming conflicts. Especially considering the fact that pretty much every meaningful name on the npm registry is taken. I also don't really see how this would make refactoring easier.

That being said, it feels rather strange to me to have a tool called "standard" supporting the most obscure patterns ever (that being said, having multiple `node_modules` isn't very common either, therefore this PR itself was probably pointless in the first instance).

feross:
Decided to merge this because of this issue (https://github.com/feross/standard/issues/40).

ahdinosaur:
i use the `src/node_modules` pattern in my apps ([here](https://github.com/cobudget/cobudget-ui) and [here](https://github.com/holodex/app)), so the default ignore of `**/node_modules/**` doesn't work for me. what can i do?

feross:
Open to ideas about how to solve this, but please open a new issue.

PR #38, opened 2015-02-20
add --format option that uses standard-format to rewrite files

Flet:


bin/cmd.js
```diff
@@ -5,10 +5,12 @@ var standard = require('../')
 
 var argv = minimist(process.argv.slice(2), {
   alias: {
+    format: 'F',
```
feross: why not a lowercase `f`?

feross: because it's a potentially destructive operation?

Flet: Yep, exactly!

feross:
Thanks, released as 2.7.0.

PR #44, opened 2015-02-26
Add rule: Must throw an Error object

feross:
The following are considered errors by `standard`:

```js
throw "error";

throw 0;

throw undefined;

throw null;
```

The following are NOT considered errors by `standard`:

```js
throw new Error();

throw new Error("error");

var e = new Error("error");
throw e;

try {
    throw new Error("error");
} catch (e) {
    throw e;
}
```

### DIFF
--- a/rc/.eslintrc
+++ b/rc/.eslintrc
@@ -141,6 +141,7 @@
         "no-sync": 0,
         "no-ternary": 0,
         "no-trailing-spaces": 2,
+        "no-throw-literal": 2,
         "no-undef": 2,
         "no-undef-init": 2,
         "no-undefined": 0,


feross:
This could just be a minor version bump since it doesn't cause any of the modules in `test.js` to fail.

mafintosh:
:+1:

mafintosh:
the wording in this PR confused me a bit since we are talking about errors when using errors

mafintosh:
would this still work?

``` js
var notFound = function() {
  var err = new Error('not found')
  err.status = 404
  return err
}

throw notFound()
```

feross:
@mafintosh Good point about the wording, fixed! :)

feross:
@mafintosh, yep that example would work, except you forgot the space after `function` ;)

dcousens:
To be clear, this PR essentially only stops the throwing of non-`Error` literals?

feross:
Yes. You can technically do "throw myFunc()" where myFunc() returns a
string. This rule prevents the obvious cases where someone tries to throw a
string or other literal that's not an Error

PR #48, opened 2015-03-01
Make example run standard after tests have passed

motdotla:
Loving standard, @feross. Even the lack of semi-colons.

The only thing I found in my flow - especially switching to standard was:

I prefer to see format run AFTER my tests. This way, I'm focused on getting code to pass first and format second. 

This pull request just updates the README to reflect that personal opinion.

Sooo, this is an extremely nitpicky and mostly personal pull request, but I thought the majority might have a similar flow of programming thought. :smile_cat:

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ First, install `standard`. Then, install the appropriate plugin for your editor:
       "standard": "*"
     },
     "scripts": {
-      "test": "standard && node my-normal-tests.js"
+      "test": "node my-normal-tests.js && standard"
     }
   }
   ```


dcousens:
This is just an example though?  No?

motdotla:
Exactly. Just an example. 




A lot of people will likely just copy and paste the example. I'd argue it'd be better they copy and paste with it at the end. 




Like I said, super nitpicky here and one of the least worthwhile pull requests I've done. But it's late and I'm at a hackathon so maybe not the clearest thinking either right now.

feross:
Thanks for the PR. I actually like having `standard` run first because it only takes 1-2 seconds, while the rest of my tests frequently take 1-2 minutes (remote browser tests are slow). I'd rather fail early than late.

motdotla:
Cool cool.

PR #52, opened 2015-03-02
add standard style badge

maxogden:
I made a badge that people can use in their readmes:

[js-standard-style](https://github.com/feross/standard)

Also if someone wants to re-implement in SVG later on they should send another PR


feross:
NICE!

PR #53, opened 2015-03-03
ESLint no-dupe-args rule

mightyiam:
Not yet released.
https://github.com/eslint/eslint/commit/64cdf9e8ad73032d7831dd96ecb539cfe1e1d485.

### DIFF
--- a/rc/.eslintrc
+++ b/rc/.eslintrc
@@ -76,6 +76,7 @@
         "no-debugger": 2,
         "no-delete-var": 2,
         "no-div-regex": 0,
+        "no-dupe-args": 2,
         "no-dupe-keys": 2,
         "no-else-return": 0,
         "no-empty": 0,


feross:
What happens if you include a rule that is not recognized? Should we wait
until release before merging?

mightyiam:
I don't know, @feross, so I would just wait for next release and check whether the rule is indeed in it.

Are you following releases?

feross:
Yep, I'm bumping the eslint dep anytime I notice it's out of date. If you notice a release and we haven't updated yet, please comment here or open a new issue.

Flet:
Looks like travis answered your question: throws error if rule is not found. :)

feross:
This PR is included in `standard@3.0.0-beta`.

mightyiam:
Thank you!

PR #54, opened 2015-03-03
Add atom plugin to readme

ricardofbarros:
Hello, sorry for the shameless plug but I think this can be very useful for Atom users that want to use your standard.


maxogden:
nice, I just tried it out and it is pretty cool, thanks!

feross:
Yeah, thanks for sharing!!

PR #63, opened 2015-03-12
make --format work for stdin

Flet:
This PR will make stdin work when using the --format option. It will also lint the formatted version of stdin which mimics the behavior of --format on files.

All feedback welcome, thanks!

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -31,8 +31,7 @@ if (argv.help) {
       -v, --verbose    Show error codes (so you can ignore specific rules)
           --stdin      Force processing input from stdin
           --version    Display the current version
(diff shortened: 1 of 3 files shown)


Flet:
Also, I see a failure in tests for `parse-torrent`

```
parse-torrent: Error: Use JavaScript Standard Style (https://github.com/feross/standard)
  bin/cmd.js:27:6: file is already defined (eslint/no-redeclare)
not ok
```

Not sure if this is due to recent rc changes or out-of-date formatting on `parse-torrent`.

feross:
The `parse-torrent` error doesn't exist in the `v3` branch, so let's ignore it

feross:
Released as 2.11.0.

Flet:
Thanks!

Issue #61, opened 2015-03-11
support app node_modules


ahdinosaur:
we use the [`src/node_modules` pattern](https://medium.com/node-js-javascript/working-without-frameworks-part-1-b948f281f782) in our apps (e.g. [here](https://github.com/cobudget/cobudget-ui) and [here](https://github.com/holodex/app)), so the default ignore of `**/node_modules/**` doesn't work for us.

previous discussion in https://github.com/feross/standard/pull/35.

possible solutions already brought up:
1. ["we could ignore node_modules if a package.json is present in the same directory"](https://github.com/feross/standard/pull/35#issuecomment-74837466)
2. [implement a way to un-ignore a directory](https://github.com/feross/standard/pull/35#issuecomment-74775195)

1 makes a lot of sense, i don't see how [it wouldn't work](https://github.com/feross/standard/pull/35#issuecomment-74993242) as the `src/node_modules` pattern as i understand it doesn't involve a `src/package.json`. keen to hear what people think, am happy to implement a PR for this.

feross:
Yeah, I actually think 1 would work. Happy to consider a PR for it.

PR #66, opened 2015-03-17
support app node_modules

ahdinosaur:
i implemented option 1 from #61: "we could ignore node_modules if a package.json is present in the same directory".

rough sketch of changes (feedback appreciated):
- removed the `**/node_modules/**` ignore pattern
- added `**/package.json` to the search patterns and also to the ignore patterns
- before ignore, iterate over list of files matched by search pattern and if it is a `package.json`, add `join(basename(file), 'node_modules/**)'` to the ignore patterns
- change order of de-duping files and regexifying ignore patterns

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,7 +24,7 @@ var ESLINT_REPORTER = path.join(__dirname, 'lib', 'eslint-reporter.js')
 var ESLINT_REPORTER_VERBOSE = path.join(__dirname, 'lib', 'eslint-reporter-verbose.js')
 
 var DEFAULT_IGNORE = [
-  '**/node_modules/**',
+  '**/package.json',
   '.git/**',
   '**/*.min.js',
   '**/bundle.js',
@@ -80,12 +80,10 @@ function standard (opts) {
       ? opts.files
       : [ '**/*.js' ]
 
+    patterns.push('**/package.json')
+
     // traverse filesystem
     if (opts.ignore) ignore = ignore.concat(opts.ignore)
-
-    ignore = ignore.map(function (pattern) {
-      return new Minimatch(pattern)
-    })
 
     parallel(patterns.map(function (pattern) {
       return function (cb) {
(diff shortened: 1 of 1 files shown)


feross:
I just merged branch `v3` into `master` and published `v3.0.0`. Can you rebase and update this PR? I simplified the way that `glob` is used here: https://github.com/feross/standard/commit/b03bcc34dbc23efbc0d2becae88f09e29b12a006 so your code will require some changes.

feross:
Hey @ahdinosaur, if you still want to work on fixing up this PR, please now rebase against the `speed` branch, as that will become the new `master` in a few days and it makes significant changes.

ahdinosaur:
@feross sweet, thanks for the heads up. i'm still keen to fix this PR, just considering the best approach to do it now with the new and improved code, mostly due to the lack of separate search and filter steps which i relied on.

feross:
Hey Mikey :) It was nice to meet you in NZ!

ahdinosaur:
@feross likewise, i'm really glad to be able to associate a human to the bits and pixels. thanks for being awesome. :smile_cat:

regarding this PR, i'm stalling on it because the only solution i can imagine feels so hacky and wrong: do a pre-search to find `package.json` files so you can add those directories' `node_modules` to the list of ignores.

i'm not sure how best to resolve this. it seems like the only sensible way to handle the app modules pattern without crazy hacks is to by default only ignore `/node_modules`, and maybe support [other ignores](https://github.com/feross/standard/pull/35) through https://github.com/feross/standard/issues/95 and the `package.json` fields, but i'm not sure if that's acceptable for this project.

ngoldman:
My 2 cents: reading #61 and #95 it seems like the best option is to respect `.gitignore` and let it do the work of sorting out what should and shouldn't be ignored by `standard` beyond root `node_modules`.

feross:
Yeah, I think respecting `.gitignore` makes the most sense. PR out: https://github.com/feross/standard/pull/107

feross:
Closing this now, since .gitignore support is released in 3.4.0, and the default `**/node_modules/**` ignore was changed to `node_modules/**`.

ahdinosaur:
sweet, thanks @feross!

feross:
No problem! :-)

Issue #75, opened 2015-03-22
importScripts incorrectly reported


JakeChampion:
```
// worker.js
importScripts('../../pouchdb.min.js')
```

If you run this file through `standard`, it returns 

``` bash
worker.js:2:0: 'importScripts' is not defined.
```

`importScripts` is a valid function in a worker context and as such, shouldn't error.

JakeChampion:
`self` is also another property that exists in workers that `standard` returns a false positive for.

feross:
This edge case is documented in the readme now: https://github.com/feross/standard#what-about-web-workers

The recommendation is to just add a comment to the top of your worker:

```js
/* global self, importScripts */
```

Issue #11, opened 2015-01-28
web workers


calvinmetcalf:
the global variable in a web worker is named self, so renaming this to self in a worker can be tricky if you then need to grab a reference to the global, does this assume it's always in a browserify context?

feross:
For web workers, I would just add this to the top of the file:

```
/* global self */
```

I'm hesitant to do this by default because then we lose the protection of checking for accidental use of `window.self` when people accidentally forget `var self = this` in a method.

feross:
Closing this now, since adding a comment is a pretty nice solution :)

PR #76, opened 2015-03-22
added self, importScripts and close to globals

reg4in:
Battles the problem described in #75.

### DIFF
--- a/rc/.eslintrc
+++ b/rc/.eslintrc
@@ -16,7 +16,11 @@
         "escape": false,
         "navigator": false,
         "unescape": false,
-        "window": false
+        "window": false,
+        
+        "self": false,
+        "close": false,
+        "importScripts": false
     },
 
     "plugins": [


JakeChampion:
I think this may be a bug in eslint.

``` JS
// worker.js
importScripts('../../pouchdb.min.js')
const db = new PouchDB('example')
```
If you run this file through `standard`, it returns 
``` BASH
worker.js:3:0: 'PouchDB' is not defined.
```

calvinmetcalf:
so there are 2 related issues, the first is that standard would like you to always use the global object when referring to globals, aka in the browser you'd go window.foo = PouchDB

BUT in web workers the global object is named self which conflicts with the rule which says that you always have to do `var self = this` 

but tl;dr from #11 do 

```js
//worker.js
/* global self */
self.importScripts('../../pouchdb.min.js')
const db = new self.PouchDB('example')
```

reg4in:
Oops forgot about that `var self = this` thing ...

feross:
Yeah, this is a tough one. We want to prevent accidental use of the browser's unfortunately named `window.self` variable, which is why `self` is not added as a global.

Right now, I recommend you add `/*global self */` to the top of web workers and service workers to fix the `standard` errors and be extra explicit that you expect this file to be run as a Worker. I agree with you that this is not ideal, but I think it's the best we can do at the moment. We can revisit this later when/if there's a better solution.

Thanks for opening the original issue and sending this PR :)

maxogden:
I documented this in the readme https://github.com/feross/standard/blob/master/README.md#what-about-web-workers

feross:
Thanks @maxogden!

PR #84, opened 2015-03-25
Add example git pre-commit hook to the readme

brentburg:


brentburg:
This is useful for large projects that may not want to update their entire code base to pass the standard code style all at once. It only holds committers accountable for the files they have contributed/changed.

maxogden:
cool idea, thanks

maxogden:
I moved it down into the FAQ

PR #90, opened 2015-03-27
fix cwd bug when using as module

Flet:
I think I found a bug when trying to use `standard` programatically... I included a little test file to illustrate my problem and also a proposed solution.


Flet:
Am I way wrong on this?

feross:
You're right, this was an issue. Fixed in 3.3.1.

PR #92, opened 2015-03-27
Add vim configuration.

santiagogil:
If you have a regular syntastic install with javascript plugins
included, this is needed to specify the use of standard.


yoshuawuyts:
:100:, this is indeed needed.

feross:
Thanks!

santiagogil:
Thank you.
Love the simplicity of standard.

PR #100, opened 2015-04-02
Improve size of images.

Yeti-or:


maxogden:
@Yeti-or can you describe these changes?

Yeti-or:
I reduce size of png image with ImageOptim.app & ImageAlpha.app and size of
svg image with svgo.
Let's save some traffic, For people who use these badges.

feross:
LGTM. Thanks!

PR #107, opened 2015-04-08
Ignore files in .gitignore

### LINKED ISSUE #61: support app node_modules

### LINKED ISSUE #95: ignore files in .gitignore
i sometimes have an *experimental* folder in a project where i try out different things. i usually add that to .gitignore. standard however still checks javascript files inside but since that folder sometimes contains some crazy bad code it usually fails. this makes my tests not run locally because of `standard && tape test/*.js`.

it would be great it standard just ignored files ignored by .gitignore (similar to how npm works).

feross:
I think it makes sense to ignore the files in `.gitignore` by default. Anything that's not checked into git (generated files, `node_modules`, temporary files, etc.) shouldn't be getting linted. I don't see a downside to ignoring them. This should hopefully decrease the need for people to use the `"standard"` property of `package.json`.

Fixes #95 - support .gitignore.
Fixes #61 - support the "app node_modules pattern". Replaces PR #66.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,8 +149,9 @@ This module saves you time in two ways:
 
 ### How do I ignore files?
 
-The paths `node_modules/`, `.git/`, `*.min.js`, `bundle.js`, and `coverage/` are automatically excluded
-when looking for `.js` files to style check.
+The paths `node_modules/**`, `*.min.js`, `bundle.js`, `coverage/**`, and hidden
+files/folders (beginning with `.`) are automatically excluded when looking for `.js` files
+to style check.
 
 Sometimes you need to ignore additional folders or specific minfied files. To do that, add
 a `standard.ignore` property to `package.json`:
--- a/index.js
+++ b/index.js
@@ -2,7 +2,6 @@ module.exports.lintText = lintText
 module.exports.lintFiles = lintFiles
 
 var dezalgo = require('dezalgo')
-var dotignore = require('dotignore')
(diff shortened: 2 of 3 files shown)


feross:
Ideally, we should fix https://github.com/bmeck/dotignore/issues/3 before we merge this, since the handling of ignore patterns differs slightly from git at the moment.

Collaborators, please hold off on merging this for a little bit ;)

feross:
Released as 3.4.0.

PR #113, opened 2015-04-14
Use stdout instead of stderr

### LINKED ISSUE #112: stdout instead of stderr
Currently standard uses stderr for all output. This means that if you want to pipe standard, you need to redirect stderr to stdout with `2>&1 >/dev/null`.

All lint errors should instead be printed to stdout so unix pipes work.

The general semantics of stdout is "content I'm interested in if I'm running this program" and stderr is "any content that is not related to the functionality of this program".  i.e. these lines are appropriate for stderr: 
https://github.com/feross/standard/blob/master/bin/cmd.js#L89
https://github.com/feross/standard/blob/master/bin/cmd.js#L108

feross:
Reasoning behind this change:
- https://github.com/feross/standard/issues/112#issuecomment-92544684
- http://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout

Fixes #112.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,15 +256,6 @@ No. Use `eslint` directly if you want to configure hundreds of options individua
 Pro tip: Just use `standard` and move on. There are actual real problems that you could
 spend your time solving! :P
 
-### Why can't I pipe standard to other tools?
-
-`standard` prints to `stderr`. This means that tools that read from `stdout` won't be
-able to read its output. The solution is to make `standard` print to `stdout` instead:
-
-```sh
-standard 2>&1 | grep variable
-```
-
 ### What about Web Workers?
 
 Web workers have a magic global variable called `self`. In regular JS files, `standard`
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -83,17 +83,17 @@ if (argv.stdin) {
 }
 
 function onResult (err, result) {
-  if (err) return error(err)
+  if (err) return onError(err)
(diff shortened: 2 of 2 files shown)


malandrew:
lgtm!

malandrew:
One quick question: Any reason why you need to put it into verbose mode to get the rule? It would be nice if that was also part of the standard mode since that makes building a checkstyle or json linter no longer dependent on the --verbose flag.

feross:
@malandrew That's a different issue from this PR. In short, the vast majority of the time you're going to just fix the lint errors. Overriding an eslint rule is supposed to be extremely rare, so I think it makes sense to hide the eslint rule name behind `--verbose` in order to keep the output simple. Feel free to open an issue if you want to solicit other opinions.

malandrew:
After I finish write this reporter module, I'll look into that issue. I believe it is only necessary for the checkstyle reporter, since the rule name would be put in the `source` xml property.

My main reason for giving people the name of the rule, is so they can go look up the documentation on that rule to better understand it. 

i.e.: print rule name next to each error and in verbose mode, at the end of linting, print a deduped list of URLs for the documentation of each violated rule so someone can go read the docs and better understand a rule.

But anyways, that's a different issue and I'll open an issue later. This PR is good to go.

Flet:
:+1:

feross:
Released as 3.6.0.

PR #118, opened 2015-04-19
Add section about Git pre-commit hook

lencioni:
overcommit is a fully configurable and extendable Git hook manager that
includes out-of-the-box support for running standard as a Git pre-commit
hook.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,6 +281,16 @@ git diff --name-only --cached --relative | grep '\.js$' | xargs standard
 exit $?
 ```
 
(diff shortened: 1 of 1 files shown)


feross:
There's already a section about git pre-commit hooks at the end of the readme. Can you shorten this description up and move it there?

lencioni:
Ah! I passed right over that. Yes, that makes much more sense. Done! Let me know if you'd like any other changes.

feross:
thanks for the contribution!

PR #127, opened 2015-05-03
Fixed a bug in parseOpts that causes default ignore patterns to be added to original opts object

leomelin:
I found this causing problems in [gulp-standard](https://github.com/emgeee/gulp-standard) where lintText -function is being used in a loop causing opts object to contain exponential times the ignore patterns slowing tasks down and eventually causing crash.

### DIFF
--- a/index.js
+++ b/index.js
@@ -112,6 +112,9 @@ function lintFiles (files, opts, cb) {
 function parseOpts (opts) {
   if (!opts) opts = {}
 
+  // Clone opts -object to make sure the original opts object is not modified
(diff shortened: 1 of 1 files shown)


feross:
thanks!

feross:
We set a function property on the options object to make the command line `--format` command work, so I removed the JSON.stringify/parse and replaced with xtend. Let me know if that doesn't fix your issue for some reason.

https://github.com/feross/standard/commit/23b475cf73dbef3a051ded2df68b93f790bc4716

feross:
Released as 3.7.3.

PR #147, opened 2015-05-29
streamline `package.json` checking with `pkg-config`

ahmadnassri:
I've seen a number of methods for using `package.json` for dependencies' options/configs, so I created [`pkg-config`](https://github.com/ahmadnassri/pkg-config) to streamline access.

I actually used your code / lookup method as the basis for the package as it made the most sense and was the cleanest, added some improvements to `pkg-config` and opted to nest the options under `config` in keeping with npm's standard with allowing to change the root (as in this PR's implementation)


ahmadnassri:
FYI, test failure are un-related to changes here ...

feross:
Thanks for the PR. Can you rebase this on top of the latest changes? Also, I have other feedback inline.

index.js
```diff
@@ -4,7 +4,7 @@ module.exports.lintFiles = lintFiles
 var dezalgo = require('dezalgo')
 var eslint = require('eslint')
 var extend = require('xtend')
-var findRoot = require('find-root')
+var config = require('pkg-config')('standard', { root: false })
```
feross: We need to pass in `opts.cwd`, otherwise this is going to introduce a new bug.

index.js
```diff
@@ -119,25 +119,14 @@ function parseOpts (opts) {
   // Add user ignore patterns to default ignore patterns
   var ignore = (opts.ignore || []).concat(DEFAULT_IGNORE_PATTERNS)
 
-  var root
-  try {
-    root = findRoot(opts.cwd)
```
feross: This is where we used to pass in `opts.cwd`. But that's no longer used with your changes. Please fix.

ahmadnassri:
reabased & updated.

thanks!

feross:
Why did you re-add the `find-root` dependency? I thought the point was to replace that with the higher-level pkg-config?

ahmadnassri:
you're still using it for reading `.gitignore`

https://github.com/feross/standard/blob/master/index.js#L166

feross:
Released as 4.2.1.

ahmadnassri:
awesome, keep up the good work! I've already converted all company projects to use `standard`

PR #165, opened 2015-06-18
test: add more repos to clone test

yoshuawuyts:
:tada: Thanks!
## Changes
- __test__: add @yoshuawuyts repo's to `test/clone`

### DIFF
--- a/test/clone.js
+++ b/test/clone.js
@@ -132,7 +132,24 @@ var URLS = [
   'https://github.com/ngoldman/module-init.git',
   'https://github.com/ngoldman/wireframe.css.git',
   'https://github.com/npm/fstream.git',
-  'https://github.com/othiym23/packard.git'
+  'https://github.com/othiym23/packard.git',
+  'https://github.com/yoshuawuyts/assert-npm-version.git',
+  'https://github.com/yoshuawuyts/brick-router.git',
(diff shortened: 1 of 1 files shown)


feross:
Thanks!

dcousens:
@feross will there come a point where we have *too* many?

feross:
Never! ;-)

I would test against every module that has `standard` in their package.json "test" script field if that was feasible. ;-) It's a way of catching backwards-incompatible changes before they get released, even if it's only advisory.

The main argument against adding more clone tests is the time it takes to run the full suite. I just pushed a change that parallelizes the cloning from git and running `standard`. It also runs `git pull` instead of `git clone` on subsequent test runs, to save time/bandwidth. https://github.com/feross/standard/commit/95faa06a49ca13e0ba9f7cc5ec236be08caaaa7d

This reduced test run time from 3:50 before, to :55 on first run and :45 on subsequent runs! :+1: 

I also added a "test-quick" script that runs in just 1 second if you're developing and just want a quick sanity check.

feross:
Just got it down to ~:38 on first and subsequent runs with https://github.com/feross/standard/commit/d5600edbd77775b407528ffcaaa217320afc7374

Flet:
This is great! :sparkles:

PR #168, opened 2015-06-22
only use relative paths in ignore check. fixes #166

### LINKED ISSUE #166: Problem when using the full path when ignoring files
I have a folder named `~/code/node_modules` in which I put the source code of all the modules I'm developing. But as of v4.0.0 it seems that standard will now use the full path of a file when choosing to ignore it or not (I think it was introduced by 647672192bb7188b8255be7302211acbb7e9f582). And since my `.gitignore` is set up to ignore the folder `node_modules`, standard will now ignore all my files since the parent two levels up is also called `node_modules`.

I think it would be an easy fix to just use the relative path, but I'm not sure so I would like to hear your comments :)

mafintosh:
I have the same problem as described in #166. PR fixes that by using relative paths when running through the gitignore filter. Note that two tests are failing but I also had the same two tests fail on master.

### DIFF
--- a/index.js
+++ b/index.js
@@ -103,7 +103,7 @@ function lintFiles (files, opts, cb) {
 
     if (opts._gitignore) {
       if (os.platform() === 'win32') files = files.map(toUnix)
-      files = opts._gitignore.filter(files)
+      files = toAbsolute(opts.cwd, opts._gitignore.filter(toRelative(opts.cwd, files)))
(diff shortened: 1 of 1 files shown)


feross:
Yeah, test failures sometimes happens with a test suite where you're cloning and testing other people's code :)

Thanks for fixing this! :+1:

feross:
Released as 4.3.2.

PR #173, opened 2015-06-25
Use https:// link for eslint tarball instead of git://

achingbrain:
git:// tries to use ssh behind the scenes which is usually blocked by corporate networks.  https:// tends not to be so use that instead.

This prevents this sort of error:

```
npm ERR! git fetch -a origin (git://github.com/eslint/eslint.git) fatal: unable to connect to github.com:
npm ERR! git fetch -a origin (git://github.com/eslint/eslint.git) github.com[0: 192.30.252.131]: errno=Operation timed out
npm ERR! Error: Command failed: fatal: unable to connect to github.com:
npm ERR! github.com[0: 192.30.252.131]: errno=Operation timed out
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   },
   "dependencies": {
     "dezalgo": "^1.0.1",
-    "eslint": "git://github.com/eslint/eslint.git#3f5cf79d65560ed69ff29b11d85047554c4d1ae1",
+    "eslint": "https://github.com/eslint/eslint/tarball/3f5cf79d65560ed69ff29b11d85047554c4d1ae1",
     "eslint-config-standard": "3.3.0",
     "eslint-config-standard-react": "1.0.0",
     "eslint-plugin-react": "^2.1.0",


Flet:
Indeed, I have this same problem :)

I've updated my `git` config to rewrite git URLs to https and my overall experience with npm and git behind a firewall has improved a lot :)

```
git config --global url."https://".insteadOf git://
```
There is a great [stack overflow](http://stackoverflow.com/questions/4891527/git-protocol-blocked-by-company-how-can-i-get-around-that/10729634#10729634) answer that covers the details really well.

That being said, this change is certainly nice to folks who can only use https, however there is no guarantee that upstream dependencies will not use git URLs. 

With npm improving git support over the past couple months, I suspect we will see even more of this in the ecosystem.

feross:
Sorry about that – it's my fault for being overeager to get some eslint fixes in and introducing the github dependency. We should remove that as soon as eslint does a release (which should be any day now since it's been 2 weeks since the last release).

feross:
Released as 4.3.3.

feross:
I just released 4.4.0 which depends on eslint 0.24, so this should no longer be an issue.

PR #187, opened 2015-07-13
use deglob to reconcile file list

Flet:
Hello!

I extracted a new package, `deglob`, from the existing code in `standard`. Its goal is to take a list of glob patterns and return an array of file locations, while respecting `.gitignore` and allowing for ignore patterns via `package.json`.

I wanted to use the same ignore/gitignore behavior in `standard-format` and I also thought this behavior would be useful for other CLI tools. 

To ensure no bottleneck, I have added all collaborators of `standard` to deglob (and gave them publish access in npm).

Feedback on `deglob` itself is welcome! :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,18 +1,15 @@
 module.exports.lintText = lintText
 module.exports.lintFiles = lintFiles
 
+var deglob = require('deglob')
 var dezalgo = require('dezalgo')
 var eslint = require('eslint')
 var extend = require('xtend')
 var findRoot = require('find-root')
 var fs = require('fs')
-var glob = require('glob')
-var ignorePkg = require('ignore')
 var os = require('os')
-var parallel = require('run-parallel')
 var path = require('path')
 var pkgConfig = require('pkg-config')
-var uniq = require('uniq')
 
 var DEFAULT_PATTERNS = [
   '**/*.js',
@@ -30,6 +27,12 @@ var DEFAULT_CONFIG = {
   configFile: path.join(__dirname, 'rc', '.eslintrc'),
   reset: true,
   useEslintrc: false
+}
+
+var DEGLOB_OPTIONS = {
+  useGitIgnore: true,
+  usePackageJson: true,
+  configKey: 'standard'
 }
 
 /**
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -78,67 +81,34 @@ function lintFiles (files, opts, cb) {
...
-  opts._gitignore = ignorePkg()
-
-  function addIgnorePattern (patterns) {
-    opts._ignore = opts._ignore.concat(patterns)
-    opts._gitignore.addPattern(patterns)
-  }
+  var ignore = DEFAULT_IGNORE_PATTERNS.slice(0) // passed into glob
 
-  if (opts.ignore) addIgnorePattern(opts.ignore)
-  if (opts.parser) useCustomParser(opts.parser)
```
feross: I think you accidentally removed this line. We still want to support custom parsers!

package.json
```diff
@@ -13,7 +13,8 @@
   "bugs": {
     "url": "https://github.com/feross/standard/issues"
   },
-  "dependencies": {
+"dependencies": {
```
feross: Indentation is weird here

index.js
```diff
@@ -78,67 +81,34 @@ function lintFiles (files, opts, cb) {
...
     }
     return cb(null, result)
   })
+
 }
 
 function parseOpts (opts) {
   if (!opts) opts = {}
-  opts = extend(opts)
+  opts = extend(opts, DEGLOB_OPTIONS)
```
feross: Can you pass the options for `deglob` directly into that function rather than merging them in with the ones that standard requires? I think that's clearer. Remember to pass through options that are used by both, like `cwd` and `ignore`.

feross:
Comments are inline. Otherwise looks great~

Flet:
Thanks! I've address the inline comments. Are the deglob options more clear now?

feross:
Looks good! Released as 4.5.4.

(Tests were failing because eslint got stricter and started catching more `no-redeclare` cases. Hopefully the packages in our tests will fix their code as they start to update `standard`)

Flet:
:sparkles: :sparkles: woo :)

Yeah I noticed those failures as well... At least they failed the same way before/after :)

PR #189, opened 2015-07-16
feat(rules) add eslint-plugin-standard

xjamundx:
- [x] add `eslint-plugin-standard`

### DIFF
--- a/rc/.eslintrc
+++ b/rc/.eslintrc
@@ -1,3 +1,10 @@
 {
-  "extends": ["standard", "standard-react"]
+  "extends": ["standard", "standard-react"],
+  "rules": {
+    "standard/object-curly-even-spacing": [2, "either"],
+    "standard/array-bracket-even-spacing": [2, "either"]
+  },
+  "plugins": [
+    "standard"
+  ]
 }
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "eslint-config-standard": "4.0.0-1",
     "eslint-config-standard-react": "1.0.2",
     "eslint-plugin-react": "^2.1.0",
+    "eslint-plugin-standard": "^1.0.0",
     "find-root": "^0.1.1",
     "get-stdin": "^4.0.1",
     "minimist": "^1.1.0",


rc/.eslintrc
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": ["standard", "standard-react"]
+  "extends": ["standard", "standard-react"],
+  "rules": {
+    "object-curly-even-spacing": [2, "either"]
```
xjamundx: Not really able to add this custom rule to the public configuration plugins, for obvious reasons.....may be easier when these custom rules become their own plugin at some point in the future.

index.js
```diff
@@ -25,6 +25,8 @@ var DEFAULT_IGNORE_PATTERNS = [
 
 var DEFAULT_CONFIG = {
   configFile: path.join(__dirname, 'rc', '.eslintrc'),
+  reset: true,
```
feross: Isn't this not necessary anymore with eslint 1.0.0 since reset is the default behavior, i.e. there are no default rules? I just removed this in https://github.com/feross/standard/commit/4f7b20d621441e8ee2c60a375c112a56a3d64e75 ;)

xjamundx: Oh sorry that came in during the rebase :)

rules/object-curly-even-spacing.js
```diff
@@ -0,0 +1,264 @@
...
+      validateBraceSpacing(node, first, second, penultimate, last)
+    }
+
+  }
+
+}
+
+module.exports.schema = [
+  {
+     'enum': ['always', 'never', 'either']
```
feross: can you fix the indentation here? it's causing the tests to fail.

xjamundx:
Seems all the github repos aren't downloading or something? Any ideas what's wrong?

index.js
```diff
@@ -25,6 +25,7 @@ var DEFAULT_IGNORE_PATTERNS = [
 
 var DEFAULT_CONFIG = {
   configFile: path.join(__dirname, 'rc', '.eslintrc'),
+  rulePaths: ['rules'],
```
feross: This should be `rulePaths: [ path.join(__dirname, 'rules') ]`

feross:
Yeah, the path for `rules` needs to be absolute.

xjamundx:
```
# tests 131
# pass  116
# fail  15
```

feross:
The solution is to take a look at each individually and decide if it's acceptable to make people change their code when they upgrade to standard v5. In cases where their spacing is actually inconsistent, I think that's fine (and I'll comment out those repos after I merge your PR).

But it's also catching things like:

```
var target = new DW({ Directory: true, type: 'Directory',
                      path: __dirname + '/output'})
```

Does it make sense to check when the object goes across multiple lines? What does the default rule do?

feross:
Also, it's catching this:

```js
var arr = [
  { str: "var x = ['test.js', '**test/failing/**']\n",
    msg: 'Noop on singleline arrays'
  },
  { str: 'function x () {}\n',
    msg: 'Noop on named functions correctly spaced'
  }
]
```

Thoughts?

xjamundx:
I thought I fixed those types of cases. Let me me check it out

xjamundx:
Still trying to figure out this is catching: 
```js
var target = new DW({ Directory: true, type: 'Directory',
                      path: __dirname + '/output'})
```

Either way. I'll beef up the rule soon and get it back up here after taking into account these cases. thanks for caching them!

xjamundx:
Okay so far all the ones I saw that were caught were legit!

xjamundx:
Will start to work on the array-bracket-spacing rule now

feross:
Awesome!

xjamundx:
Okay 2nd rule is in.

xjamundx:
I didn't do `computed-property-spacing`. Let me know if that's needed. I figured you might be better off just enforcing no space for that. It's super rare to see spaces in that.

xjamundx:
Also if you do feel like using a plugin for this stuff. Both of the rules live here as well:
https://github.com/xjamundx/eslint-plugin-standard

feross:
I think making a new package makes a lot of sense. Can you add the current GitHub collaborators so they auto-watch all the issues on that repo?

feross
maxogden
mafintosh
othiym23
dcposch
Flet
dcousens 
jprichardson

Agreed about the `computed-property-spacing` rule.

feross:
Actually, if it's not too hard, could you implement `computed-property-even-spacing`? It's not done often, but some of our dependents do do that I can't blame them:

```js

  fs[ self.props.follow ? 'stat' : 'lstat' ](p, function (er, stat) {
    if (er) return self.error(er)
    ...
  })
```

feross:
Also, could we change the `There should be no space before X.` message to make it clearer that we're not enforcing removing spaces, but only evenness? Here's an idea:

`Use balanced spacing - there should no space before X."

Thoughts?

xjamundx:
Sure, I'll turn this issue into an issue integrating the plugin. I'll publish it but add  you as an owner in npm soyou can publish updates etc

Flet:
Great work! :)
Separate plugin is ideal to me as it means integration with `semistandard` will be smooth (and `standard-engine` can stay rule agnostic).

xjamundx:
Check out this updated PR

dcousens:
LGTM well done @xjamundx :)

rc/.eslintrc
```diff
@@ -1,3 +1,10 @@
 {
-  "extends": ["standard", "standard-react"]
+  "extends": ["standard", "standard-react"],
+  "rules": {
+    "standard/object-curly-even-spacing": [2, "either"],
```
xjamundx: We could put these in the `eslint-config-standard` but then other would have to `require('eslint-plugin-standard')` to use it without it blowing up. Might be nice to get feedback from consumers of `eslint-config-standard` such as @Flet

dcousens: Agreed,  as a user of `eslint-config-standard` also,  I still don't know the answer.  @Flet thoughts?

Flet: I think adding the rules to `eslint-config-standard` would work similar to how [eslint-config-standard-react](https://github.com/feross/eslint-config-standard-react) works.

`eslint-standard-react` uses `eslint-plugin-react`, which is specified in `peerDependencies` on that package. So, when `eslint-standard-react` is installed, npm automatically installs the plugin too. (https://docs.npmjs.com/files/package.json#peerdependencies)

So, adding this plugin to `eslint-config-standard` should be OK as long as the plugin is added to `peerDependencies`.

xjamundx: :+1:

xjamundx: https://github.com/feross/eslint-config-standard/pull/5

feross: **Note:** npm3 will stop automatically installing `peerDependencies`. [Read more.](https://github.com/npm/npm/blob/master/CHANGELOG.md#peerdependencies) and only show a warning when the peerDependency is not installed by the parent user.

PR #201, opened 2015-07-21
Add global options to command line

LeiZeng:


feross:
Thanks for the PR! I simplified things a bit and merged this.

Released as 5.0.0-3. This is a prerelease version, so beware that breaking changes may be added before 5.0.0 is released.

PR #202, opened 2015-07-21
Allow using ESLint's formatters

SimenB:
There's probably a valid reason for this not being a thing, but as it's an easy change, I threw together this PR anyways :smile: 

Basically, I prefer the default (stylish) output, and not the boring one included in this package.

Another approach would be to have it as a flag, and not a string. Then only allowing either the included formatter, or eslint default (stylish atm).

Need to add it to package.json config as well (and docs), but as that wasn't immediately apparent to me how to accomplish, I'll wait for this to be an accepted change. No reason wasting time on it if you're not interested in this :smile:

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -5,6 +5,7 @@ var minimist = require('minimist')
 var standard = require('../')
 var standardFormat = require('standard-format')
 var stdin = require('get-stdin')
+var eslint = require('eslint')
 
 var argv = minimist(process.argv.slice(2), {
   alias: {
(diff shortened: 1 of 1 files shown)


feross:
I would prefer to solve this outside of `standard` core, since it's not an essential feature.

I published a new package [`snazzy`](https://github.com/feross/snazzy) that does just what you need: it transforms "compact" output to "stylish".

after

Compared to before:

before

## install

```
npm install -g snazzy
```

## usage

Pipe "compact" text into the `snazzy` command to get back pretty results:

```bash
$ standard --verbose | snazzy
```

Or, just run `snazzy` directly and it will run `standard` and give you pretty results:

```bash
$ snazzy
```

`snazzy` supports all command line flags that `standard` supports:

```bash
$ snazzy --format --verbose test1.js test2.js
```

SimenB:
Sweet, I'll use that. Thanks!

PR #204, opened 2015-07-22
fix global cli

### LINKED ISSUE #200: Global variables options in command line.
Hi, @feross ,

I know we can set options in package.json or put some comments in the head of the files.
But we still met some troubles when use some plugins in our Editors(like Atom), the cwd is no longer the root path of our project, so if we could have some options to config the global variables in the command line like the ignores, it will be perfect!!

ricardofbarros:
This will fix #200

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -26,8 +26,6 @@ var argv = minimist(process.argv.slice(2), {
   ]
 })
 
-console.log(argv)
-
 // running `standard -` is equivalent to `standard --stdin`
 if (argv._[0] === '-') {
   argv.stdin = true
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -20,9 +20,13 @@ var argv = minimist(process.argv.slice(2), {
     'version'
   ],
   string: [
-    'parser'
+    'parser',
+    'global',
+    'globals'
   ]
 })
+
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -110,7 +110,7 @@ function parseOpts (opts) {
   if (!opts.ignore) opts.ignore = []
   opts.ignore = opts.ignore.concat(DEFAULT_IGNORE)
 
-  setGlobals(opts.globals || opts.global)
+  setGlobals(opts.global)
```
feross: This change is not right. I want to support either option name when people `require('standard')` in node.js. Even if the command line uses only `global`, we should keep the other.

bin/cmd.js
```diff
@@ -85,6 +87,7 @@ if (argv.stdin) {
       })
     }
   }
+
```
feross: Avoid random whitespace changes next time

feross:
Thanks! Released, with fixes as described, as 5.0.0-4.

bin/cmd.js
ricardofbarros: My bad

index.js
ricardofbarros: Okay, understood :sweat_smile:

Issue #87, opened 2015-03-26
Suggestion: splitting core out


mattdesl:
Just flowing some thoughts here, let me know if you guys are interested or not.

What if the core was split into its own module to better facilitate forks like `semistandard` and create an ecosystem like `google-standard`, `airbnb-standard`, etc. The module would just handle ESLint reporting with a given config. Then `standard` would be a thin wrapper around that module, using the current `.eslintrc`. 

Having everybody on the same page is great for npm as a whole, but also a difficult goal to attain given the diversity of JS devs and their work. For example; a lot of client work I do gets touched by many devs (freelancers, juniors, co-op students, etc) and runs on very tight deadlines. Rules like `newline at end of file`, `space after comments`, `single quotes only` slow down our development and lead to distracting lint noise.

However, it would still be beneficial for us to have a tool like `jam3-standard` so that our modules can catch true syntax errors. It also means that we can releases changes to our code style using SemVer, which is (in my view) a huge innovation brought forward by `standard`.

Flet:
A nice command line wrapper around an .eslintrc file is an interesting idea.

I don't know if letting a thousand standards bloom fits with the initial motivations of creating this tool. An all purpose wrapper may require some broader design decisions, such as bringing in additional linters.

I guess all it takes is someone motivated enough to fork and work on it!

jprichardson:
I'm against this unless it somehow benefits `standard`.

Creating a `google-standard`, `airbnb-standard`, etc are antithetical to the goal of `standard`. Fortunately `standard` is OSS, so if someone really wants to, they could easily clone it and create `google-standard` and friends similar to how `semistandard` was created. 

> Having everybody on the same page is great for npm as a whole, but also a difficult goal to ascertain given the diversity of JS devs and their work. For example; a lot of client work I do gets touched by many devs (freelancers, juniors, co-op students, etc) and runs on very tight deadlines. Rules like newline at end of file, space after comments, single quotes only slow down our development and lead to distracting lint noise.

Totally understand what your concern here. My suggestion would be to just create simple `eslint` rules that suit you and your team's development and expand from there.

mattdesl:
@jprichardson 

The benefits: separation of concerns, modularity and semantic versioning. e.g. The `eslint-error-stream` module could be versioned independently of the `rc` file that `standard` uses. If `standard` decides to drop ESLint, it would be trivial to just remove it as a dependency.

Forking `standard` is bound to lag behind; it's a burden to constantly pull in patches.

[...]

PR #213, opened 2015-07-30
switch to standard-engine

Flet:
This PR will switch `standard` to use `standard-engine`.

`standard-engine` is a package I've been keeping synced up with `standard`. It includes only the "engine" bits and none of the rules. We're using it currently on `semistandard` to keep it synced up without a lot of hassle. Its also used on `doublestandard` too :)

This PR points to a prerelease version of `standard-engine@2.0.0-1` which is using `eslint@1.0.0-rc-1`, so it should work out nicely to be part of `5.0.0` if thats possible.

There are some slight differences in the CLI `--help` text. We can change `standard-engine` if folks wish.

All collaborators from here have been added to `standard-engine` and are free to treat it with the same care as `standard` :) Also, all git commits from `standard` were preserved, so contributions should still be reflected there too.

Anyone have thoughts on this?


Flet:
This should also close #87 if its accepted

Flet:
...just realized deps need tobbe cleaned up now

dcousens:
I've used standard-engine myself in https://github.com/JedWatson/happiness,  so I'm pro this idea.

Flet:
Synced up `standard` with `standard-engine@2` and picked apart `cli.js` and `index.js` and got them much closer to `standard` now. 

Dependencies should be a lot better now.

Flet:
Sweet! Now I guess I should make the `standard-engine` README a little better...

feross:
Thanks, this is now merged! I made a few fixes in `standard-engine` and re-locked the dependencies that you loosed in `package.json`. It's a lot easier to debug issues that come up when dependencies aren't shifting. And we're currently doing pretty regular releases, so I'm not worried about people not getting bugfixes :)

Flet:
Excellent! :+1: 

Thanks! Makes sense to lock things down. And thanks for finding those bits I missed.

feross:
No problem. Really happy that we can share the effort here and you don't have to manually keep things in sync anymore.

feross:
Released as 5.0.1.

PR #218, opened 2015-08-03
Use the built-in ESLint reporter (Stylish)

kasperisager:
This PR refactors Standard to report errors using the built-in reporter in ESLint.

Before:

screen shot 2015-08-03 at 19 46 37

After:

screen shot 2015-08-03 at 19 49 23

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -94,15 +94,7 @@ function onResult (err, result) {
     '(https://github.com/feross/standard)'
   )
 
-  result.results.forEach(function (result) {
-    result.messages.forEach(function (message) {
-      log(
-        '  %s:%d:%d: %s%s',
-        result.filePath, message.line || 0, message.column || 0, message.message,
-        argv.verbose ? ' (' + message.ruleId + ')' : ''
-      )
-    })
-  })
+  log(result._getFormatter()(result.results))
 
   process.exit(1)
 }
--- a/index.js
+++ b/index.js
@@ -47,11 +47,13 @@ function lintText (text, opts, cb) {
   cb = dezalgo(cb)
 
   var result
+  var engine = new eslint.CLIEngine(opts._config)
   try {
-    result = new eslint.CLIEngine(opts._config).executeOnText(text)
+    result = engine.executeOnText(text)
   } catch (err) {
     return cb(err)
   }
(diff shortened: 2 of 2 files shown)


jprichardson:
Not sure what @feross thinks about this PR, but presumably that's why he built [snazzy](https://github.com/feross/snazzy).

Flet:
Unfortunately, I don't think this will get merged...
>"... the output format of standard is locked – it will never change" 
https://github.com/feross/standard/issues/142#issuecomment-111284255

And:
>"I would prefer to solve this outside of standard core, since it's not an essential feature.
I published a new package snazzy that does just what you need: it transforms "compact" output to "stylish"."
https://github.com/feross/standard/pull/202

I definitely recommend checking out [snazzy](https://github.com/feross/snazzy)!

kasperisager:
Snazzy it is then!

feross:
I'm afraid this change will break the editor tools that rely on the current output format on stdout.

We could try to detect if standard is running in a TTY (with `process.stdin.isTTY`) and use the stylish formatting in that case, falling back to the current formatting in scripts and when piping data, but this method is not foolproof.

LinusU:
I would love to see some nicer output from standard, but I don't think that the extra effort (although small) of using snazzy outweighs that need.

I'm +1 on this getting merged thought, maybe with `isTTY`. But somehow I think that we have kind of the worst of two worlds currently. It's not very human readable and it's not very machine readable. I don't know, just my two cents...

jprichardson:
It's best to just keep it the same. You can never predict how consumers of your software will use it. `isTTY` feels like a "non-pure" hack. Mocha uses this and it forced me to do some nasty things in `[electron-mocha](https://github.com/jprichardson/electron-mocha)`... but that's because they probably didn't anticipate a Node environment that's DOM-enabled.

feross:
Thanks for the PR, but we don't need this feature in core.

There are many of modules that style the output - it's not that hard to parse. The built-in reporter works just fine, but there's also [snazzy](https://www.npmjs.com/package/snazzy), [standard-json](https://www.npmjs.com/package/standard-json), and [standard-tap](https://www.npmjs.com/package/standard-tap) for specific styles. Also [standard-reporter](https://www.npmjs.com/package/standard-reporter) supports a bunch more.

PR #224, opened 2015-08-05
Minor update on Readme.md

eenagy:
babel-eslint should be installed globally


jprichardson:
But you may not want to install `babel-eslint` globally?

feross:
You only need to install `babel-eslint` globally when you're using `standard` globally. Can you update the note? Then I'll merge this.

feross:
I took care of it. Thanks again for the PR.

PR #235, opened 2015-08-26
fence js / json

reggi:
Edited some blocks so that it passes evalMd (https://github.com/reggi/evalmd). Example objects should be js blocks with variable declared.


reggi:
I don't understand all I did was update the `README.md`. 

@feross the next level is to ensure the the code within a README.md is valid, it's to ensure that it's standard:

```
evalmd README.md -so='concat' | standard --stdin
```

feross:
@reggi - there was a bug in the tests. Not your fault.

Merged.

PR #237, opened 2015-08-27
Remove duped word from tagline message

keis:
"Use" is added by standard-engine resulting in the following tagline being printed

```standard: Use Use JavaScript Standard Style (https://github.com/feross/standard)```


dcousens:
Thanks @keis

PR #243, opened 2015-09-03
State that callback is not optional

tjwudi:
callbacks are generally optional, but not here though. It might be confusing for somebody.

### DIFF
--- a/README.md
+++ b/README.md
@@ -454,6 +454,8 @@ var results = {
 }
 ```
 
+Note that `callback` is not optional.
+
 ### `standard.lintFiles(files, [opts], callback)`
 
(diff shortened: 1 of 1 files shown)


dcousens:
> callbacks are generally optional

That is an interesting assumption.  I'd always assume they are not optional unless explicitly noted.

feross:
> callbacks are generally optional

Not sure where you're getting this from. In node.js core, callbacks are usually required when the whole point of the call is to get back some result. If you omit the callback, then there's no point in calling `standard.lintText` since you won't learn anything.

Also, you can tell which parameters are optional by looking for the square brackets. For example: `fn(required, [optional])`

See: `standard.lintText(text, [opts], callback)`

PR #252, opened 2015-09-08
Fix "globals" property typo

febeling:
In the example package.json, the property name wasn't correct.


feross:
Both "global" and "globals" will actually work, but okay, sure. :)

PR #254, opened 2015-09-08
Update README.md

ricardofbarros:
`linter-js-standard` doesn't need `linter` installed beforehand as prior. [`59cbc84`](https://github.com/ricardofbarros/linter-js-standard/commit/59cbc84b6e2eb9ab3ec9aac50b77ff23d4d8c9a3)


feross:
Thanks for letting us know -- readme updated!

PR #258, opened 2015-09-16
standard should run after other test scripts

stevemao:
You might want to just quickly test but you don't care about stylings for now. For example, you add a `console.log` or you add some scripts with no indentation, and then you wanna quickly run the test scripts.


feross:
Feel free to make this change in your own projects, but I prefer the example the way it is now. Thanks for the suggestion.

PR #274, opened 2015-09-25
Update link to repos that use standard

mnquintana:
Just realized that there wasn't actually a list of repos in https://github.com/feross/standard/blob/master/test/clone.js - this updates the link to point to the list of repos! :beers:

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ npm install standard
 To get a better idea, take a look at
 [a sample file](https://github.com/feross/bittorrent-dht/blob/master/client.js) written
 in JavaScript Standard Style, or check out some of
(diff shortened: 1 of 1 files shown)


Flet:
Cool, good idea! Perhaps it should point to [standard.json](https://github.com/feross/standard-packages/blob/master/standard.json) as this is a bigger, more definitive list?

Also, we should totally make this list web-friendly (in an automatied way) and add it to the web site on `standard-web`.

mnquintana:
>Cool, good idea! Perhaps it should point to standard.json as this is a bigger, more definitive list?

Ah, not sure how I missed that one - done!

> Also, we should totally make this list web-friendly (in an automatied way) and add it to the web site on standard-web.

That would be awesome :+1:

feross:
:+1:

PR #276, opened 2015-09-29
Add noote about multiline ternary operator rules

rstacruz:
Takes care of:

> `'?' should be placed at the beginning of the line. (operator-linebreak)`


feross:
thanks!

PR #280, opened 2015-10-02
link this awesome collection of Atom snippets

capaj:


Flet:
Ah, did not know this existed, neat! It should probably be added to [feross/awesome-standard](https://github.com/feross/awesome-standard) too.

feross:
Merged.

PR #286, opened 2015-10-09
Link standard badges to standardjs.com rather than github

sonnyp:


feross:
good suggestion, thanks!

PR #291, opened 2015-10-12
switch to win-spawn for windows compatibility

Flet:
Switch to using `win-spawn` in `clone.js` to ensure Windows compatibility.

https://github.com/ForbesLindesay/win-spawn

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "run-parallel-limit": "^1.0.1",
     "standard-packages": "^1.2.0",
     "tape": "^4.0.0",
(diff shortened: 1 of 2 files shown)


jprichardson:
utACK :+1:

Flet:
Tested on Windows with node@4.2.0 and both the `clone` and `pull` scenarios are working. 

Its incredible how slow it runs on Windows though. For just a `pull` of each repo, It took about 5 minutes on Windows and 1 minute on an Ubuntu VM running on that same Windows OS :open_mouth:.

feross:
Nice work!

PR #302, opened 2015-10-23
Rules: add note about var assignments

rstacruz:
documents the `one-var` rule

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -189,6 +189,21 @@ your code.
     'www.api.com'
   ```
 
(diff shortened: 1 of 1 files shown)


Flet:
:+1: 

We may also want to point out that its OK to put uninitialized vars on the same line:

```js
// ✓ ok
var x, y, z
```

dcousens:
LGTM, can't merge until CI passes

feross:
Passing now, thanks to @flet :)

feross:
merged.

PR #314, opened 2015-11-01
remove "?style=flat" in badge link

fscherwi:
#### Remove "?style=flat" in badge link, because it's unnecessary!

| style | badge |
|-----------------------|----------------------------------------------------------------------------------------|
| with "?style=flat" | []() |
| without "?style=flat" | []() |


yoshuawuyts:
Oh yeah, I believe it used to be `flat-square` so it might be residue from when someone changed that.

Flet:
Thanks!

feross:
Flat didn't used to be the default style, but glad it is now :)

PR #323, opened 2015-11-09
README.md: suggest installing version `*`

mightyiam:
I remember that it was the recommended way in the past.

Why not anymore?

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,8 @@ Search the extension registry for **["Standard Code Style"][brackets-1]**.
     }
   }
   ```
+  
+  Or set the version to `"*"` to encourage using the latest version.
 
 2. Check style automatically when you run `npm test`
 


watson:
It was changed in 9b35dcb76d2166693562e1650390408ba354d963 which doesn't really help much in explaining. I guess it's just good practice in case something breaking is added, though in case of standard it could be argued you'd always want the latest.

dcousens:
For large projects, this is a good idea.
For small projects where the SLOC is low, and you probably just want travis to occasionally remind that something might be wrong,  `*` is probably good advice.

mightyiam:
I argue suggest both. Because many may not even think of going `*`, where they would end up trying it, at least.

yoshuawuyts:
If people know what they're doing (e.g. are familiar with npm), `*` is a fine choice for some situations. If you don't know what you're doing and are following the instructions in the `README.md` for the first time ever it might be best to suggest something more stable.

mightyiam:
@yoshuawuyts familiarity with npm, or any tool, with most people, is not complete. For example, I learned about the `*` version symbol from this one right here. And it proved to be useful.

I don't remember learning about it since then, so who knows if I'd have ever thought of putting the `*` for use in this way.

mightyiam:
I argue: suggest it in some way. If what I've written isn't clever enough, could you suggest some other wording?

yoshuawuyts:
Generally you should never use `*` as it can cause things to break between builds / requires patches to do more fixes than strictly necessary. I feel as if this PR is slightly bikesheddy as it shouldn't be `standard`'s task to educate people on what npm can and cannot do. I'm :-1: on this PR.

rstacruz:
:-1: This will eventually break something in someone else's build. Use something like [npm-check](https://www.npmjs.com/package/npm-check) or [greenkeeper](http://greenkeeper.io/) instead.

yoshuawuyts:
Closing this issue as it seems the general concensus is that promoting a `*` version is often a very bad idea. If anyone has a strong opinion otherwise, you're welcome to reopen this PR to continue the discussion.

@mightyiam Even though this isn't being merged as of now, please don't let this disencourage you from suggestion other changes whenever you run into something. Your contributions are much appreciated :sparkles:

PR #325, opened 2015-11-13
Follow Standard in Examples

skorlir:
It's a tiny thing, but it bugs me that the examples at the very end don't follow the standard! I propose the above correction. I don't think there's anything else wrong.

Let me know what you think!


dcousens:
Thanks @skorlir

PR #361, opened 2015-12-15
standard-format@2.1.0

LoicMahieu:
`standard-format@2.1.0` is shipped with `esformatter-jsx@3` including `babylon@6`


feross:
I suggest you use `standard-format` directly since we actually plan to remove it from `standard` in v6 which will be released soon. See discussion here: https://github.com/feross/standard/issues/340

PR #368, opened 2016-01-02
Correct atom package name

DaleJefferson:
The name of the snippets package in atom is standardjs-snippets not atom-standardjs-snippets


feross:
Thanks!

Issue #289, opened 2015-10-11
Never start a line with ` because of tagged template strings


gyson:
because

```js
let x = () => {}

`hello`

console.log(x)
```

will output 

```
undefined
```

(tested with nodejs v4.1.0)

rstacruz:
Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings#Tagged_template_strings

feross:
Issue opened on eslint: https://github.com/eslint/eslint/issues/4169

feross:
Turns out this is a bug in V8. See the bug I filed here: https://code.google.com/p/v8/issues/detail?id=4497

A semicolon should automatically be inserted by the JS engine. The fact that it isn't in node v4 is a bug. Fortunately, the V8 developers say it's fixed in `master` so it'll make it's way to node soon enough.

Mystery solved.

gyson:
Thanks for your clarification.

Just one more question:

Would this case
```js
let x = () => {}
let y = x
`hello`
```
be parsed to
```js
let x = () => {};
let y = x`hello`;
```
or
```js
let x = () => {};
let y = x;
`hello`;
```
?

feross:
@gyson I think it will actually be parsed to the first one. :( Fortunately, the eslint developers are aware and plan to improve the rule: https://github.com/eslint/eslint/issues/4169#issuecomment-149064446

gyson:
@feross I see. Thanks.

PR #376, opened 2016-01-08
Rules: add semicolons before template strings

rstacruz:
#289 — maybe we should add this, what do you think?

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -227,12 +227,16 @@ your code.
   window.alert('hi');  // ✗ avoid
   ```
 
-* Never start a line with `(` or `[`. This is the only gotcha with omitting semicolons.
+* Never start a line with `(`, `[` or `````. This is the only gotcha with omitting semicolons.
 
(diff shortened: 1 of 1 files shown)


bcomnes:
Looks good to me for the same reasons not to start lines with [ and ( 

Just to clarify, this is still necessary unrelated the V8 bug? https://bugs.chromium.org/p/v8/issues/detail?id=4497

dcousens:
LGTM.

It should probably be re-iterated that these rules **are enforced** by `standard` and as such you will never run into a situation where this is actually a problem.

These cases so rarely appear in production code,  its quite moot as an issue.

feross:
@rstacruz Thanks for this PR, but it doesn't cleanly merge anymore. I manually merged it into the `v6` branch.

Here it is: https://github.com/feross/standard/commit/0a9c24fc17e804f0c8d34967ed185b0c3f9e16c4

PR #391, opened 2016-01-26
Improve the description of 'no semicolons'

rajcoss:
> This is the only gotcha with omitting semicolons.

This statement seems to convey that its still a gotcha, and programmers need to keep this is mind even while using standard for code linting. However, since standard catches and flags it, its not a gotcha anymore.


dcousens:
LGTM :+1:

PR #397, opened 2016-02-01
Remove standard-format. fixes #340

### LINKED ISSUE #340: Move standard --format into a separate module
Currently around ~1/3 of the dependencies are there to support `standard --format`. I propose we move that command to a separate module `standard-format` since it isn't related to linting and would make standard a lot "lighter" to install.

mafintosh:
With this PR standard has around 250 fewer dependencies when installing.

### DIFF
--- a/options.js
+++ b/options.js
@@ -9,7 +9,5 @@ module.exports = {
   tagline: 'Use JavaScript Standard Style',
   eslintConfig: {
     configFile: path.join(__dirname, 'rc', '.eslintrc')
-  },
-  formatter: require('standard-format'),
-  formatterName: 'standard-format'
+  }
 }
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
haha, lgtm :sparkles:

dcousens:
@mafintosh does this break `standard --format`?

feross:
@dcousens Yes, it will. We're going to recommend that folks install `standard-format` separately from now on. The dependencies for `standard-format` are just too many and it makes install take way too long, even more-so with npm 3.

feross:
@mafintosh This looks good, but I want to print a friendlier message to people who try to run `standard --format`. Maybe something like:

```
Code formatting is available in a standalone package: "standard-format". To install it, run "npm install -g standard-format".
```

dcousens:
ACK :+1:

feross:
This has been merged into the `v6` branch.

Issue #371, opened 2016-01-06
Add "env" config to package.json


gsklee:
As a continuation of #18, can we add this?

```json 
{
  "standard": {
    "env": "mocha"
  }
}
```

IMHO it's better than either adding `/* eslint-env mocha */` to all test files, or adding:

```json 
{
  "standard": {
    "globals": ["beforeEach", "describe", "it", ...]
  }
}
```

dcousens:
IMHO, if `/* eslint-env mocha */` is a thing, than this is exactly what should be added to all test files.
I don't think blanket application of a `mocha` env to an entire project is acceptable unless it is 100% applicable.

rstacruz:
bad, mocha. one of the reasons I don't use mocha anymore. :p

dmitriz:
@rstacruz What is better than mocha?

rstacruz:
I personally prefer tape, but that's just me.

Anyway, IMHO this is most certainly doable. Eslint itself has support for environments defined here: http://eslint.org/docs/user-guide/configuring#specifying-environments. These look particularly useful:

- browser
- worker
- mocha
- ...

rstacruz:
Also, @feross, this might be worth documenting:

```js
/* eslint-env mocha */
```

...will consider describe/it/beforeEach/etc as predefined globals.

dcousens:
I guess if it is basically just a direct pass-through for a common eslint feature, then I'm not against it.

dcousens:
Though,  IMHO `/* eslint-env mocha */` is a better pattern than `package.json` defined configuration.

rstacruz:
yes, i agree. there's also the possibility of this:

```json
{
  "scripts": {
    "lint": "standard 'lib/*.js'; standard 'test/**/*.js' --env mocha"
  }
}
```

dcousens:
@rstacruz that I would agree with a lot more than the blanket suggestion.

dmitriz:
@rstacruz Seems like they require to write this in every single test file:
> You always need to `require('tape')` in test files.

Grows old very fast, imo.

rstacruz:
dmitriz: i know this is off tangent in the conversation now, but that's precisely why I like it. It means your tests are runnable on their own (`node test/foo.js`), and perfectly browserifyable (`browserify test/index.js`). This opens up the opportunity to do a lot of things:

```sh
node test    # run in node.js
browserify test | smokestack -b firefox  # run in headless browser
[...]

PR #401, opened 2016-02-04
Readme: document mocha support

rstacruz:
I've also changed the service worker documentation, since eslint supports those natively.


rstacruz:
Related: #371

feross:
LGTM :+1:

PR #427, opened 2016-02-18
Use install location of standard as eslint `cwd`

feross:
Fixes https://github.com/feross/snazzy/issues/8


feross:
Released as 6.0.6.

PR #445, opened 2016-03-03
Update document Vim automatic formatting on save

ibotdotout:
that `standard % --format` is no longer included with standard but document about Vim automatic formatting on save not updated.


feross:
thanks!

PR #454, opened 2016-03-08
fix markdown for webstorm.md

Flet:
This fixes an issue we're running into when generating the site on standard-www


yoshuawuyts:
:100:

PR #461, opened 2016-03-16
fix markdown formatting in RULES.md

wadey:
A previous commit broke the formatting on this line: https://github.com/feross/standard/commit/e765b91c878ee454300dc3a9d76c41472d416b0c#commitcomment-16730715


feross:
@wadey Thanks!

PR #469, opened 2016-03-26
Undo config hacks added in v6.0.2 and v6.0.3

Flet:
This seems like a good thing? @feross I just rebased your branch... no other changes.

>These hacks were added to deal with bugs in eslint v2.0.0:
>
>https://github.com/feross/standard/compare/v6.0.1...v6.0.3
>
>Should not be necessary now.

### DIFF
--- a/options.js
+++ b/options.js
@@ -5,7 +5,6 @@ var pkg = require('./package.json')
 module.exports = {
   bugs: pkg.bugs.url,
   cmd: 'standard',
-  cwd: __dirname,
   eslint: eslint,
   eslintConfig: {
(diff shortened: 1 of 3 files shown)


options.js
```diff
@@ -1,26 +1,17 @@
...
-
 module.exports = {
-  eslint: eslint,
-  cmd: 'standard',
-  version: pkg.version,
-  homepage: pkg.homepage,
   bugs: pkg.bugs.url,
-  tagline: 'Use JavaScript Standard Style',
+  cmd: 'standard',
+  cwd: __dirname,
```
feross: I think this line needs to be removed.

PR #474, opened 2016-03-28
Update CLI usage when using glob patterns.

joshuajabbour:
Globs need to be quoted or they will be expanded by your shell (and not standard). This means that explicitly matching the default patterns (`standard **/*.js **/*.jsx`) will likely result in a different result than just relying on the defaults.


joshuajabbour:
See https://github.com/Flet/standard-engine/issues/84 for confusion arising from this.

feross:
Thanks!

PR #492, opened 2016-04-20
Update pre-commit hook regex and exit logic

darrhiggs:
Standard defaults to matching all files with the extension .js or .jsx
Update pre-commit hook to match files with both extensions.
Update pre-commit hook to only exit upon failure of command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,8 +407,8 @@ Funny you should ask!
 ```sh
 #!/bin/sh
 # Ensure all javascript files staged for commit pass standard code style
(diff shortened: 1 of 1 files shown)


feross:
> Update pre-commit hook to only exit upon failure of command.

What's wrong with exiting every time? Isn't it the same?

darrhiggs:
I run more than just this command in my pre-commit hook. As this is fast -
compared to my tests - I prefer to run this and only exit upon failure. I then
run more expensive commands afterwards.

feross:
Gotcha.

PR #514, opened 2016-05-07
Fix Webstorm double quoting

JBlaak:
Prevents Webstorm to format generated code with double quotes, e.g.:

`import Bar from 'bar'`

instead of

`import Bar from "bar"`


yoshuawuyts:
Looks reasonable to me

Issue #542, opened 2016-06-06
Usage in Webstorm with ES7


Alexsey:
There is a <a href="https://github.com/feross/standard/blob/master/docs/webstorm.md">guide of how to config standard for Webstorm</a> and there is a <a href="https://github.com/feross/standard#can-i-use-a-custom-js-parser-for-bleeding-edge-es6-or-es7-support">guide of how to configure it to use ES7 features</a>. Unfortunately installation of ```babel-eslint``` and mentioning it in ```package.json``` will not help Webstorm to understand ES7 syntax. Is there some other way to do that or Webstorm can't support such?

feross:
I'm not a Webstorm user, but if you figure out a solution please send a PR to update the documentation.

Alexsey:
PR is merged, answers now are in the docs

PR #545, opened 2016-06-07
Update webstorm.md

Alexsey:
#542 In Webstorm docs add description of how to enable ES7 support. Also fix numbering


Alexsey:
Sorry, there are still some issue with local install. Will create another PR when figure out what is wrong

PR #546, opened 2016-06-07
Improve Webstorm doc

Alexsey:
Add info about ES7 support, fix numeration, add info about sudo usage, add details about ESLint package selection and some other


feross:
Thanks!

PR #571, opened 2016-07-22
Fixed a couple typos.

fatpixel:
😀

### DIFF
--- a/docs/webstorm.md
+++ b/docs/webstorm.md
@@ -32,7 +32,7 @@ config settings directory
     </code_scheme>
   ```
 
-5. You may install dependencies and config globaly or localy and with support of ES7 or without it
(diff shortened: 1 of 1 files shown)


feross:
Thanks @fatpixel!

@dcousens I'll investigate the test failure. Do you generally think it's fine to enforce test passing before merge, or is that too burdensome?

dcousens:
> Do you generally think it's fine to enforce test passing before merge

I think so,  it brings it to your attention.  In other libraries I often merge with administrator rights if its something like this,  but still like how it raises my internal red flags as to "this shouldn't have to have to happen".

feross:
Yeah, this was actually good because I fixed up the tests 👍

Issue #386, opened 2016-01-15
Add eslint plugins?


jenius:
So I have been looking at [flow](http://flowtype.org/), and would very much like to try it out for a project I'm working on to get some of the lovely benefits of strong typing. However, standard isn't happy about the type annotations when I start adding them in.

I found [a nice eslint plugin](https://github.com/gcazaciuc/eslint-plugin-flowtype) that strips the type annotations for you, and started looking into whether it was possible to use this with standard, but it doesn't seem like it.

Any thoughts on the possibility of letting users add eslint plugins? In this case, the type annotations don't make it past the babel compile anyway, so it's not bending the rules of standard.js in any way. Standard already does accept custom parsers too, so I'm hoping it would be reasonable to add plugins as well.

If this does sound ok, I'd be happy to contribute!

feross:
Sure. This isn't the first issue about support for flowtype. See: https://github.com/feross/standard/issues/306, https://github.com/feross/standard/issues/186, and https://github.com/feross/standard/issues/157

I've been wanting to expose the underlying eslint configurability via `package.json` and command line option so that we can put an end to all the requests to support X, Y, or Z JavaScript extension.

In the past, I've rejected such things on the basis that we want to avoid configurability in `standard`, since that's kinda the whole point.

But, users use JSX, Babel, React, Flow, and a bunch of other things. Rather than support them in standard core, I want to move to a model where `standard` continues to Just Work out of the box, but we expose the necessary hooks for adding support for whatever flavor of the week thing decides to extend the built-in JS syntax.

So, with all that said, here's a PR to add support to `standard-engine` for setting plugins: https://github.com/Flet/standard-engine/pull/56

Works in `package.json`:

```json
{
  "standard": {
    "plugins": [ "my-plugin" ]
  }
}
```

Or via command line option:

```
standard --plugin my-plugin
```

This will be released in standard v6.

PR #604, opened 2016-08-25
Document adding eslint plugins to config

JaKXz:
@feross I couldn't find any documentation for #386, so I just did this in my config and I think it's working... just want to confirm with this :)


feross:
Updated in https://github.com/feross/standard/commit/4f6cb0e5e8589df35181432f8f398b24c933bacf

PR #619, opened 2016-09-11
Recommend `/* global */` comment over `window.`

feross:
I think using a comment at the top to make global dependencies explicit
is the nicest solution.

Context: https://twitter.com/feross/status/774766522124218368

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -149,8 +149,20 @@ your code.
   })
   ```
 
-* **Always prefix browser globals** with `window.`.<br>
-  Exceptions are: `document`, `console` and `navigator`.
+* **Declare browser globals** with a `/* global */` comment.<br>
+  Exceptions are: `window`, `document` and `navigator`.<br>
+  Prevents accidental use of poorly-named browser globals like `open`, `length`,
+  `event`, and `name`.
+
+  ```js
+  /* global alert, prompt */
+
+  alert('hi')
+  prompt('ok?')
+  ```
+
+  Explicitly referencing the function or property on `window` is okay too, though
+  such code will not run in a Worker which uses `self` instead of `window`.
 
   ```js
   window.alert('hi')   // ✓ ok
--- a/README.md
+++ b/README.md
@@ -44,9 +44,11 @@ npm install standard
 - **Space after function name** `function name (arg) { ... }`
 - Always use `===` instead of `==` – but `obj == null` is allowed to check `null || undefined`.
 - Always handle the node.js `err` function parameter
-- Always prefix browser globals with `window` – except `document` and `navigator` are okay
+- Declare browser globals with `/* global */` comment at top of file
   - Prevents accidental use of poorly-named browser globals like `open`, `length`,
     `event`, and `name`.
+  - Example: `/* global alert, prompt */`
+  - Exceptions are: `window`, `document`, and `navigator`
 - **And [more goodness][5]** – *give `standard` a try today!*
 
 [1]: http://blog.izs.me/post/2353458699/an-open-letter-to-javascript-leaders-regarding


RULES.md
```diff
@@ -149,8 +149,20 @@ your code.
   })
   ```
 
-* **Always prefix browser globals** with `window.`.<br>
-  Exceptions are: `document`, `console` and `navigator`.
+* **Declare browser globals** with a `/* global */` comment.<br>
+  Exceptions are: `window`, `document` and `navigator`.<br>
+  Prevents accidental use of poorly-named browser globals like `open`, `length`,
+  `event`, and `name`.
```
nolanlawson: I'd remove "poorly-named" in order to sound less judgmental.

Also fun fact: any element with an `id` also creates a global variable! Great source of nasty bugs. 😛

yoshuawuyts:
I feel this approach is inferior to using `window` because:
- `/* */` style flags feel like a form of config, despite being localized. Though for obvious purposes it should be allowed, making it recommended doesn't feel clean
- it encourages copy-pasting of boilerplate; I have seen this happen with ESLint before, and don't think `standard` would be different here
- on large files there can be a disparity between the declarations at the top of the file, and the usage in the middle. Anytime someone drops into the middle of a file (code search, links, or otherwise) it would be unclear where certain methods come from

I hope this makes sense. Cheers!

feross:
The primary issue with the `window.` approach is that it creates code that doesn't work in Workers (web workers, service workers, shared workers, etc.) which are becoming increasingly common.

I've sent PRs to a few npm modules already so remove `window.` so they work in Workers.

Ideas to solve this?

RULES.md
feross: I noticed that in Chrome but didn't realize that it was expected, spec-compliant behavior. Wow!

LinusU:
Hmm with this in mind I'll retract my initial downvote. Maybe `global` is the way to go, but I also agree with the points that @yoshuawuyts brought up...

yoshuawuyts:
> The primary issue with the window. approach is that it creates code that doesn't work in Workers (web workers, service workers, shared workers, etc.)

lol what, I knew there was this special context thing but hadn't realized that `window` was gone. That's um, an _interesting_ decision I guess; they def find ways to keep us on our toes hey

---

> which are becoming increasingly common

Yeah, I think you're right - but I suspect that despite rising usage, workers will be an edge case for years to come. I think the way to create compatibility is the way we've done it with node, browserify and others: create the API that makes sense on top of the things we got. In this case I think the [global](https://github.com/Raynos/global/blob/master/window.js#L5) package would resolve this, as it checks the [self](https://developer.mozilla.org/en-US/docs/Web/API/Window/self) property to return the proper instance. Maybe we could suggest folks rely on this for compat?

Idk, I feel there's tradeoffs for every solution, and not a single clear "best way". In my opinion recommending explicit usage of `window` and recommending `require('global')` for compat has the best tradeoffs. Hope this was useful.

nolanlawson:
>  workers will be an edge case for years to come

Folks writing code that doesn't work in a web worker would contribute to that. ;)

IMO adding a few comments is preferable to writing code that's hostile to web/service/shared workers. (And soon compositor workers and other stuff browser vendors are cooking up.) `require('global')` seems like a fine workaround, but it's also a lot of overhead and I imagine people would prefer not to have to add another dep just to avoid bare globals.

Another reason to do bare globals is that specifying the global object has weird side effects in NW.js, where there's both a `global` and a `window` object but you have different globals hung on each one, so the only reliable thing is to do the bare global instead of explicit `global.foo` or `window.foo` (see https://github.com/pouchdb/pouchdb/issues/3229 and https://github.com/substack/node-browserify/issues/481 for discussion). (I'm not aware of any weirdness in Electron BTW because `window === global`).

> Anytime someone drops into the middle of a file (code search, links, or otherwise) it would be unclear where certain methods come from

Valid concern, especially since `global` is often abused for non-browser globals. Maybe Standard should have a directive for browser environments? I don't see Node scripts having to do e.g. `/* global Buffer, setImmediate */` so maybe browser scripts should get the same privilege.

Flet:
`env` and `globals` can be set via the `"standard"` key in `package.json` as well. This could be a good project/package specific solution. 

```js
  "standard": {
    "globals": ["window"], // eh, not required with env set to browser :)
    "env": {
      "browser": true
    }
  }
```

Does that help mitigate?

Also, I don't see the `env` feature described in the README, but it was added earlier this year via #371 (with a somewhat similar discussion).

feross:
> @yoshuawuyts it encourages copy-pasting of boilerplate; I have seen this happen with ESLint before, and don't think standard would be different here

> @yoshuawuyts on large files there can be a disparity between the declarations at the top of the file, and the usage in the middle. Anytime someone drops into the middle of a file (code search, links, or otherwise) it would be unclear where certain methods come from

Thanks for sharing these concerns. I was worried about these cases too. But I just checked: unused variables in a `/* global */` comment are treated as errors. So, every variable in a `/* global */` comment must actually be used.

feross:
@nolanlawson I would love to include browser globals by default, so users won't get warnings when they're used, but it's a much larger list: https://github.com/sindresorhus/globals/blob/07f98339930695b2267dbc815c34ee47861e9682/globals.json#L162-L778 Over 600!

In the past, I've considered adding just the ones that begin with capital letters, since they're less likely to be used accidentally. e.g. `WebSocket` and `RTCPeerConnection` vs. `top`, `length`, `offsetHeight`. But it's still quite a long list.

Open to other ideas.

feross:
I do think the best way forward is to recommend `/* global var */` comments over `window.var` since I don't want Workers to be second-class citizens.

But I don't think that `standard` should enforce this. `window.var` works just fine and should continue to be allowed.

Flet:
Since there is no actual rules changing, is this safe to merge?

Does it make sense to just explain that either `/* global */` or `window` can be used, but web workers is a special case?

Merge timer has started! This will be merged by Friday if no one comments! ⏱ 😁

feross:
This has been sitting here for a long time. I'm going to merge it but I'm happy to see anyone send an improved wording.

Really, I think both the `window.alert` and `/* global alert */` approach should be discussed so the user can choose the best option for their circumstances.

PR #621, opened 2016-09-14
Explain how to call standard if installing locally

dandv:


dcousens:
If you're using `npm`,  you can just do:

``` json
"scripts": {
  "standard": "standard"
},
```

Then just do `npm run standard`.

timoxley:
Also see https://www.npmjs.com/package/npm-run

dcousens:
Alternatively, if you like to live a little dangerous.

> export PATH=$PATH:$NODE_PATH/bin:./node_modules/.bin/

feross:
@dandv Thanks for the PR. Could you give some background about your reasons for this change?

It seems like you're trying to make it clearer where the `standard` binary gets installed? Is there some way that we can make this clearer without specifically advocating that folks start invoking `standard` with `./node_modules/.bin/standard`?

dandv:
@feross: That was my reasoning. I prefer to install as few modules globally as possible (avoids having to sudo too).

Perhaps it's better to advocate adding standard to the `test` suite, as @dcousens suggested?

timoxley:
@feross perhaps the "What you might do if you're clever" content of the readme should simply be moved up into the "usage" section.

feross:
Fixed by @timoxley's PR #624

PR #624, opened 2016-09-14
Add example of usage within npm scripts

timoxley:
Related https://github.com/feross/standard/pull/621#issuecomment-247145748

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ in JavaScript Standard Style, or check out some of
 
 - [Install](#install)
 - [Usage](#usage)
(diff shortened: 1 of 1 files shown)


feross:
Thanks for sending this, but I liked the old wording ("what you might do if you're clever"), even though it's technically less informative in the ToC. And also the list format made the steps clearer.

timoxley:
@feross Updated. Moved it as-is.

feross:
Thanks!! Sorry for being a nitpick ;)

timoxley:
not at all an issue.

PR #627, opened 2016-09-16
Update dependencies

feross:
Upgrade eslint from 3.3.1 to 3.5.0, which adds 5-10 new rules that are fixable with `--fix`.

There were also new rule additions. I will create separate issues for the ones that make sense.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "url": "https://github.com/feross/standard/issues"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


feross:
Test failure can be ignored, it's related to a rule catching more errors in eslint 3.5.0.

feross:
Re-running tests...

feross:
Thanks for the review, @LinusU!

This is now released as `standard` v8.1.0.

PR #644, opened 2016-09-29
update vscode editor extension part

chenxsan:
I have forked a new [vscode-standardjs](https://github.com/chenxsan/vscode-standardjs) from [vscode-eslint](https://github.com/Microsoft/vscode-eslint) to support standard v8.x.x.


feross:
@chenxsan Thanks for updating to support the new `standard --fix`.

Can you update this PR to remove the reference to `vscode-standard-format`? IMO, it would be better to just change the first line to say: "Install `vscode-standardjs` (includes support for automatic formatting)"

PR #653, opened 2016-10-10
Better alt text for standard JS badges

robinwhittleton:
The existing ‘JavaScript Style Guide’ is a bit generic and doesn’t really define what standard does. This amends the alt text to be more explanatory to the user, and to better reflect the contents of the image.

This also updates the alt text of the README header to match the text in the image.


feross:
I appreciate what you're trying to do here, so thanks for trying to improve this. But the phrase "JavaScript Style Guide" is intentionally part of the alt text. I want to get http://standardjs.com to the first page of Google results for that term. We're in position 11 right now, so very close.

At the same time, accurately describing the purpose of the image is a good idea too. How about we go with "Standard - JavaScript Style Guide"

robinwhittleton:
Works for me, I’ve pushed that update.

PR #666, opened 2016-10-25
Readme: Suggest Unicode 2019 apostrophe in strings

### LINKED ISSUE #662: Use typographic apostrophe
> Single quotes for strings – except to avoid escaping

If you’d use the typographic apostrophe you wouldn’t have to make this exemption.

Typewriter apostrophe: 'we\'d have to escape'
Typographic apostrophe: 'we don’t have to escape'

Typographic apostrophe:
Mac: Shift + Alt + #
Linux: Alt Gr + #
Windows: Alt + 0146
HTML: `&rsquo;`

See also:
https://en.wikipedia.org/wiki/Apostrophe#Computing
http://practicaltypography.com/apostrophes.html
http://typefacts.com/artikel/apostroph

mightyiam:
Fixes #662

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,8 @@ npm install standard
 ### The Rules
 
 - **2 spaces** – for indentation
-- **Single quotes for strings** – except to avoid escaping
+- **Single quotes for strings** – except to avoid escaping –
+  but, by all means, use ’ (Unicode 2019) in your strings
 - **No unused variables** – this one catches *tons* of bugs!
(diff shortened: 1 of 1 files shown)


mightyiam:
I'd be happy if there's agreement that another suggestion of wording is better.

feross:
This is a neat trick, but I think this is too much info to point out at the top of the readme.

mightyiam:
:sweat: You’re a tough one, @feross...

PR #667, opened 2016-10-25
readme: suggest cleverer npm scripts integration

mightyiam:


mightyiam:
This is what I did so far and I love it.

jprichardson:
This is also the convention I have adopted as well. This way I can run `lint` on Travis by itself.

dcousens:
Concept ACK, this is also what I do

feross:
I think either way is fine, but for the readme example I prefer it the current (simpler) way.

PR #705, opened 2016-11-25
Updated configuration file for WebStorm

prigara:
`<XML>` is not required in the configuration.
`<option name="SPACE_WITHIN_BRACKETS" value="true" />` contradicts the code style used in the samples;
`var list = [ 1, 2, 3, 4 ]` instead of `var list = [1, 2, 3, 4]`


prigara:
Option `Spaces within Object literal braces` adds spaces that are not present in the [example file ](https://github.com/feross/bittorrent-dht/blob/master/client.js#L100).

PR #715, opened 2016-12-09
Exit with code 0 if node less than 4

fanatid:
or other code, but not zero

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -2,6 +2,7 @@
 
 if (process.version.match(/v(\d+)\./)[1] < 4) {
   console.error('standard: Node v4 or greater is required. `standard` did not run.')
+  process.exit(1)
 } else {
   var opts = require('../options')
   require('standard-engine').cli(opts)


LinusU:
This will break CI for a lot of people, why do you need this?

fanatid:
Yes it break CI, because this it should be merged to next major release.
Right now I don't need this anywhere, but I think that current behaviour is wrong.

dcousens:
> Right now I don't need this anywhere, but I think that current behaviour is wrong.

Why though?

feross:
I agree that it feels a bit weird to return a passing exit code on old versions of node. But this was an intentional decision. We do this because of the way that most people use `standard`, i.e. running it on multiple versions of node in CI. More info here: https://github.com/feross/standard/pull/590

Returning a failure exit code doesn't actually help a user on Node 0.10 or 0.12 at all. They can't do anything to fix the issues, short of upgrading, so better to let it pass.

PR #739, opened 2017-01-07
Added --save-dev to readme

lukasoppermann:
I think it would make sense to add the `--save-dev` flag, since this would mean `standard` is installed when people clone and `npm install` a project. Since one advantage is that teams have a common base, this seems to be a typical use case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ This module saves you (and others!) time in two ways:
 Install with:
 
 ```
-npm install standard
+npm install standard --save-dev
 ```
 
 ### The Rules


LinusU:
The build is failing because of https://github.com/sebdeckers/express-history-api-fallback/pull/8

LinusU:
Also, very small nit, but I personally prefer `--save-dev standard` (the flag before), and this is what I've seen in most other projects...

feross:
We already cover `--save-dev` and `--global` a few paragraphs later in the readme [here](https://github.com/feross/standard/blob/master/README.md#install), but this is the second time someone's opened a PR like this, so okay, sure 🚢

PR #748, opened 2017-01-14
add readme & rules in spanish

falmar:
I've in the need to share this repository/package with my coworkers and few people of which some of them does not speak English. And I though would be nice to make this contribution. So this repo could reach by more people in Spanish spoken community.

I appreciate any feedback.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
   <a href="https://www.npmjs.com/package/standard"><img src="https://img.shields.io/npm/v/standard.svg" alt="npm version"></a>
   <br>
   Readme Translation: <a href="docs/README-ptbr.md">Português</a>
+  Readme Translation: <a href="docs/README-esla.md">Spanish</a>
 </p>
 <br>
 
--- a/docs/README-esla.md
+++ b/docs/README-esla.md
@@ -1,0 +1,607 @@
+<h1 align="center">
+  <a href="http://standardjs.com"><img src="https://cdn.rawgit.com/feross/standard/master/sticker.svg" alt="Standard - JavaScript Style Guide" width="200"></a>
+  <br>
+  JavaScript Standard Style
+  <br>
+  <br>
+</h1>
+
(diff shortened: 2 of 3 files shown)


feross:
YES! This is so excellent! I'm so stoked to merge this, as well as any other language translations that people want to submit.

Thanks @falmar!

Flet:
👍 Now live on [standardjs.com](http://standardjs.com)

Flet:
OK, I made a small update to `standard-www` and now the alternate languages are live on [standardjs.com](http://standardjs.com) The site nav is still in english though :/

falmar:
Oh I see... Thanks for accepting @feross ... I'll probably give it a review again to find and fix any typo or grammar mistakes.

@Flet I just took a peek into `standard-www`. I never used the `markdown-styles` but is it possible or easy to change the nav bar according to the language or would that require a huge refactor for it?

feross:
Oh I just noticed that we're missing a link to the Spanish translation in RULES.md. I'll send a PR.

Flet:
@falmar it shouldn't be too difficult. I think there might be a more efficient way of dealing with multiple languages that makes adding new languages easier...

Maybe have each one as a sub directory like http://standardjs.com/esla/index.html

It may be easier to accomplish this if we rework the /docs directory to have a directory for each language instead (and we could include a nav file in the native language as well).

Something like: 
```
└── docs
    ├── esla
    │   ├── README.md
    │   └── RULES.md
    └── pbtr
        ├── README.md
        └── RULES.md
```

...we could also update the site generation to automatically build the links to each translation. This could be good if @feross wants to make a wider call for translations.. we could add them quickly to the site by just rebuilding.

feross:
@Flet I don't think it's too bad to maintain the inline links manually for now, especially since we're only linking back to the English version from the translations. As for the navigation links, I'm happy to do whatever you think is best.

feross:
I just tweeted about how we're looking for more translations! https://twitter.com/feross/status/822568452330291201

PR #752, opened 2017-01-20
Add translation links between language versions

feross:


feross:
Thanks for the review, @LinusU!

PR #755, opened 2017-01-21
[wip] Use curated-linter

mightyiam:
## curated-linter

https://www.npmjs.com/package/curated-linter

## Help appreciated

Any help here and in developing missing features in curated-linter will be appreciated.

## Checklist

- [ ] tests for `getConfig`
- [x] test that main export does not expose instance itself
- [ ] document breaking changes
- [ ] passing CI run
- [ ] go through readme and see what requires updating

### Drop support for Node.js v4 (see `.travis.yaml`)

This is [because of Proxies](http://kangax.github.io/compat-table/es6/#test-Proxy).

Perhaps we could use transpilation?

Do we have a support plan?

### `lintText` and `lintFiles` return promises

Time to get modern, I guess.

## Positive things to do after this merge

- [ ] extract and publish the code that tests on a list of repos (packages?), so that other curated linters may enjoy it, even though we like standard best


mightyiam:
Here is one of the failures. It is exactly in this assertion:
https://github.com/feross/standard/pull/755/files#diff-9265a69d5e81bfaab4edbd417a33a626R25

I've inserted a `console.log`, logging the `result.results[0].messages`. Seems that I'm missing something that has to do with the parsing, itself. Why is `\'c\` parsed as an identifier? Where did it even come from?

Is this the result of a difference in an ESLint version?

```
➜  standard git:(curated-linter) ✗ npm-run tape test/api.js
TAP version 13
# api: lintFiles
ok 1 result is an object
ok 2 should be equal
# api: lintText
ok 3 result is an object
not ok 4 should have used single quotes
  ---
    operator: equal
    expected: 1
    actual:   2
    at: process._tickCallback (internal/process/next_tick.js:103:7)
  ...
[ { ruleId: 'no-undef',
    severity: 2,
    message: '\'c\' is not defined.',
    line: 1,
    column: 1,
    nodeType: 'Identifier',
    source: 'c' },
  { ruleId: 'eol-last',
    severity: 2,
    message: 'Newline required at end of file but not found.',
    line: 1,
    column: 2,
    nodeType: 'Program',
    source: 'c',
    fix: { range: [Object], text: '\n' } } ]

1..4
# tests 4
# pass  3
# fail  1
```

mightyiam:
@feross I would like to be able to restart travis builds, if that's ok with you.

mightyiam:
Only 8 repos failing:
https://travis-ci.org/feross/standard/jobs/194303703#L914

At least two not due to this PR.

feross:
It seems like you put a lot of work into your package and this PR, but I don't understand what you are trying to accomplish. Next time, I recommend opening an issue to discuss your plans and get buy-in from the maintainers before you do work that might not be needed.

Specifically, dropping Node 4 is completely out of the question. It is supported until April 2018.

mightyiam:
Oh, that was a tough one.

I'm trying to write a better alternative to standard-engine. And I think I am succeeding.

What if I make it work in Node 4?

Result will be more well-defined APIs, a better architecture and far better test coverage. That's the way I see it.

And I get to be creative and contribute something significant. I'm 👍 on that.

mightyiam:
I share your approach.

I've found standard-engine a bit hard to reason about. So, while I'm reading and understanding it, I'm also rewriting it in code that is hopefully easier to understand. If you read it, you may agree.

Also, I find it to have not enough tests. I'm writing sufficient tests, in my feeling.

I'll be happy to go over all of standard-engine's code (and the bits in standard) and also the issues and make my best to do the right things. I would say that this will certainly not break the entire ecosystem. I would say that it surely will introduce some documented breaking changes, only for the API usage and not for the CLI usage. Plus some edge cases, perhaps. Good thing we have the "clone" tests, right? They should catch the vast majority of issues. And the fact that it is mostly if not completely passing already, is great, I think.

In standard-engine I've found some API decisions that I don't agree with. For example, standard-engine has some of what I would categorize as configuration. For example, the formatter, the default file patterns and the default ignore patterns. Also, the policy of not allowing `.eslintrc` files and some more things. All of that, in my mind, should be `standard` specific. And the correct abstraction is curated-linter. In that sense, it is more simple than standard-engine. It can be used to create curated linters with very different configurations, as opposed to only different rules.

Now, I understand that pulling all of those "opionions" (configuration) out of standard-engine requires putting them into standard itself. While it may seem nice for standard to be as simple as it is now, simply pushing away these opinions to standard-engine doesn't seem correct. For example, as much as we would like to take over the world (and I do desire standard to win, believe me), I think that it would be much kinder to take those opinions out of the engine, so that anyone could make his own curated linter.

Now, I don't just make the thing. I make it and implement standard with it.

I think the outcome is worth some possible edge cases breaking. And we'll have another contributing member. That's me.

feross:
> In standard-engine I've found some API decisions that I don't agree with. For example, the formatter, the default file patterns and the default ignore patterns. Also, the policy of not allowing .eslintrc files and some more things

There is no `--formatter` option anymore since we provide `--fix` by default. The default file patterns and ignores are easy to remove in a PR to `standard-engine` -- you didn't need to write a whole new package for that.

Additionally, `curated-linter` is extremely heavy compared to `standard-engine`. It's literally 40x larger and 7x slower to install on my current machine/internet connection:

```
$ slow-deps 
--------------------------------------------------------
| Dependency                 | Time  | Size   | # Deps |
--------------------------------------------------------
| curated-linter             | 24.8s | 12 MB  | 182    |
| eslint                     | 16.8s | 12 MB  | 132    |
| standard-engine            | 3.4s  | 343 KB | 24     |
```

Sorry, this makes no sense for `standard`.

mightyiam:
Before I move on, I have to know that you are not getting tired of this. I'd like to make this good and proper.

Regarding the size, there is a crucial difference that you have to be aware of while looking at the size. `standard-engine` does not depend on `eslint`. It gets it passed on to it. `curated-linter` does depend on `eslint`.

Another size issue is the AVA test runner. It is the best one out there. That's why I chose it. Yet, it depends on some babel packages. Hence a size increase. Yet, between 0.17 and 0.18 of AVA, there was a reduction of size of dependencies. So perhaps it is better now.

One other thing. `slow-deps` does not seem to document how it works, exactly. Perhaps https://www.npmjs.com/package/npm-module-stats would be a better tool to check with.

So, is there an openness with curated-linter, if I cooperate to bring it into the zone of it "making sense" for standard?

feross:
@mightyiam Thanks for pointing out that eslint is included in your package. That still leaves 50 dependencies. There are only 24 in `standard-engine`. (Only production dependencies are counted.) We also removed eslint from the engine so we could bump it separately from the engine.

Sorry, I just don't see what the point of this change is. I'm happy with `standard-engine`. I think you could continue to work on this on your own if you're still interested., or find another worthy project to start or work on.

PR #764, opened 2017-01-27
readme: Add company logos, other readme cleanup

feross:
For: https://github.com/feross/standard/issues/744

screen shot 2017-01-26 at 6 07 49 pm

Moved some of the early sections ("Badge", "Text editor plugins") into
the FAQ section, so important sections like "Why should I use
JavaScript Standard Style?" and "Who uses JavaScript Standard Style?"
come first. Once they're convinced, the can read about badges and text
editor plugins.

- Renamed, reordered some sections.
- Converted h3s into h2s since it delineates the sections better
- All logo images are under 50KB


2color:
Nice work @feross

PR #766, opened 2017-01-29
Fix link to github.com in README.md

andrew--r:


feross:
Thanks for the PR! I already merged this other one that does the same thing: https://github.com/feross/standard/pull/767 (saw it before yours)

PR #817, opened 2017-03-07
Update zhtw translation

PeterDaveHello:


PeterDaveHello:
I think "插件" is more fuzzy, it stands for not only "plugin" but also "extension" (though these two terms are also very familiar), it actually came from the mainland China, I'll say that it's a Simplified Chinese term written in Traditional Chinese.

PR #830, opened 2017-04-01
Improve pitch

feross:
1. Mention automatic fixing in the list of time-savers

2. Include the phrase "javascript style guide" in the title, which will
hopefully help with SEO. We rank 7 for that on Google right now.

3. I also think that having the phrase "JavaScript Style Guide, with
linter & automatic code fixer" appear early on in the readme is really
helpful for people to understand what `standard` is, since "One
JavaScript Style to Rule Them All" isn't really enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,15 +26,19 @@
 
 <br>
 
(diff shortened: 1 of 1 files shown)


feross:
And just like that we're #5 👍

PR #836, opened 2017-04-04
Translate RULES.md into Simplified Chinese

wayou:


feross:
Also, just noticed that the link to this translation in `RULES.md` (the English version) is missing.

PR #844, opened 2017-04-07
Print version when running

robertkowalski:
Printing the version helps when debugging issues in a team or when
running multiple standard versions.

You instantly see which version of standard is running right now.

---

image

---

no proposal written as it was a fast drive-by-change :)


feross:
Thanks for the PR!

I think the message as is doesn't read very well. Plus `standard --version` works fine, so I'm -1 on this change. But thanks anyway!

PR #851, opened 2017-04-10
Recommend w0rp/ale over syntastic

rstacruz:
https://github.com/w0rp/ale has:

- built-in support for standard (as long as `node_modules/.bin/standard` [exists](
https://github.com/w0rp/ale/blob/4b0f3257ddf4303a00979cd1171dd449bd3b9ed5/ale_linters/javascript/standard.vim)), nothing to configure.

- supports neovim.

- runs asynchronously, so it never makes your vim halt up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,7 +267,7 @@ autocmd bufwritepost *.js silent !standard --fix %
 set autoread
 ```
 
-Other linters you can try are [neomake][vim-2] and [syntastic][vim-3], both of which have built-in support for Standard (though configuration may be necessary).
+Alternative plugins to consider include [neomake][vim-2] and [syntastic][vim-3], both of which have built-in support for `standard` (though configuration may be necessary).
 
 [vim-1]: https://github.com/w0rp/ale
 [vim-2]: https://github.com/neomake/neomake
--- a/README.md
+++ b/README.md
@@ -258,11 +258,7 @@ For JS snippets, install: **[vscode-standardjs-snippets][vscode-2]**. For React 
 
 ### Vim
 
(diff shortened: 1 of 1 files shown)


rstacruz:
...of course, this is a biased opinion from a long-time vim user :) anyway, we can also recommend other linters; here's my top 3 list in order:

1. https://github.com/w0rp/ale (this)
2. https://github.com/neomake/neomake
3. https://github.com/vim-syntastic/syntastic

feross:
I like that Ale does not require configuration. 👍 That's my kind of tool! 😏

Can we add links to the others in a small comment at the bottom of the vim section? I imagine that users who already have one of the other tools installed would appreciate knowing that they already support `standard`.

yoshuawuyts:
Also a vim user here - wasn't aware of ale, looks neat!
https://github.com/w0rp/ale#5xi-will-this-plugin-eat-all-of-my-laptop-battery-power
has
me concerned though: booting and closing a node process every other
keystroke seems ummm, well a lil expensive. Perhaps this would be our cue
to start thinking about a daemon mode to negate the perf implications?
Think I'm gonna keep recommending syntastic at least for the time being.

rstacruz:
> booting and closing a node process every other keystroke seems ummm, well a lil expensive

yeah, I disabled those behaviors:

```vim
let g:ale_lint_on_text_changed=0
let g:ale_lint_on_save=1
```

while most of the time it ran fine, it sucked when working with Ruby, which did some random CPU spikes haha.

rstacruz:
> Can we add links to the others in a small comment at the bottom of the vim section?

done!

PR #894, opened 2017-05-24
Neo4j is using standard

pe4cey:
Added logo to README

Proof: https://github.com/neo4j/neo4j-browser/blob/3.0/package.json#L102


Flet:
Nice! standardjs.com should be updated shortly!

PR #898, opened 2017-05-30
Add -r flag to xargs command in git hook

### LINKED ISSUE #887: Website: git pre-commit hook should use `xargs -r`
git diff --name-only --cached --relative | grep '\.jsx\?$' | xargs -r standard
                                                                       ^^

This way, `xargs` won't run `standard` if no JavaScript files were changed in this commit. This is useful because without arguments, `standard` needlessly checks everything.

ttencate:
Fixes #887

### DIFF
--- a/README.md
+++ b/README.md
@@ -606,7 +606,7 @@ Funny you should ask!
 ```sh
 #!/bin/sh
 # Ensure all javascript files staged for commit pass standard code style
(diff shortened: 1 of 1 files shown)


ttencate:
Rats, turns out `-r` is a GNU extension. Best not force it on everybody then...

Flet:
ah I was just checking this :) Good catch

PR #900, opened 2017-05-30
Hook: prevent running standard without input files

### LINKED ISSUE #887: Website: git pre-commit hook should use `xargs -r`

ttencate:
* Update pre-commit hook to use a portable equivalent of `xargs -r`. Without `-r`, `xargs` runs `standard` without arguments, causing all files to be checked instead of none.
* Use `git diff -z`, `grep -z` and `xargs -0`, in order to handle filenames containing spaces or escape sequences.

Fixes #887

### DIFF
--- a/README.md
+++ b/README.md
@@ -603,11 +603,22 @@ $ standard --plugin html '**/*.html'
 
 Funny you should ask!
 
-```sh
-#!/bin/sh
-# Ensure all javascript files staged for commit pass standard code style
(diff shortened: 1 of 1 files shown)


Flet:
Rebased and merged with PR #1015 thanks!

Issue #908, opened 2017-06-01
Broken links inside translation documents


ematipico:
Hi team!

I was starting to work on a translation of the documentation of the project in italian.
While I was following the current translations to have an idea on how to contribute, I noticed that the links inside the foreign translations (I checked spanish, portugues and chinese so far) point to the english version.

You can try to go there https://standardjs.com/readme-esla.html and click the link "Más detalles". I haven't checked all the links - of course, but I think most of them are broken

feross:
Excited to see your Italian translation! 🇮🇹

Thanks for letting us know about the broken links. There might be a way to fix this in the build process for the website. @Flet - thoughts?

Otherwise, I think we should make a folder for each translation, `esla`, `ptbr`, `zhcn`, etc. and put the README.md and RULES.md for that language in there. That would make the relative links between README.md and RULES.md work correctly within a given translation.

We should also probably consider adding the list of translations to the header in the build script since it's a burden to update it in every translation when there's a new translation added.

PR #912, opened 2017-06-08
update link

yyjazsf:
update RULES.md link address,in zhcn file,it should be RULES-zhcn.md.
by the way,in website RULES.md can show,but in github,it doesn't work,because in no RULES.md file in at the same directory


ematipico:
#908

PR #928, opened 2017-06-21
readme: Add TypeScript instructions

feross:
TypeScript is actually supported, but most people don't know this. Add it to the readme.


feross:
Thanks for the review, @yoshuawuyts!

PR #930, opened 2017-06-22
Fix typo on JavaScript

imrobinized:


feross:
Thanks, @imrobinized!

PR #936, opened 2017-06-26
Update README.md with current repo

CafeLungo:
"feross" -> "standard". All updated static content links verified to be valid. But feel free to verify for yourself, too.

I wasn't positive on the license line if it should be changed to a standardjs.com name or not-- someone else can comment about that if it should also be updated.


feross:
> I wasn't positive on the license line if it should be changed to a standardjs.com name or not

We can leave it as copyright to me since there's not really another person or legal entity that it would make sense to put there instead.

PR #945, opened 2017-07-04
Rentograph is using standard

jaredreich:
Added logo to README

Proof: https://gitlab.com/rentograph/eslint-config-rentograph/blob/master/index.js#L2


feross:
Awesome, thanks!

feross:
The image you included is way too big (over 100KB). I will shrink and compress it.

PR #963, opened 2017-08-04
Update dependecy of eslint-config-standard-jsx

daper:
I've been having issues with a deprecation warning of `react/jsx-space-before-closing` and `eslint-config-standard-jsx` package, that are solved on `v4.0.2` by this commit standard/eslint-config-standard-jsx@c13637e.

I would like to reference this issue bcomnes/sublime-standard-format#54 too, becouse this change fixes it for me.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "dependencies": {
     "eslint": "~3.19.0",
     "eslint-config-standard": "10.2.1",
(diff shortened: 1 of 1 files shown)


feross:
Please remove the tilda character as we prefer to lock down dependencies for maximum consistency.

bcomnes:
Released as `10.0.3`.  🎉

PR #964, opened 2017-08-06
What you might do if you're cleverer?

ianwalter:
Moving standard to the "pretest" script accomplishes the same goal of having standard automatically run before "test" is run but also allows you to run it separately, e.g. `npm run pretest`.

### DIFF
--- a/docs/README-en.md
+++ b/docs/README-en.md
@@ -156,7 +156,8 @@ $ standard "src/util/**/*.js" "test/**/*.js"
       "standard": "*"
     },
     "scripts": {
-      "test": "standard && node my-tests.js"
+      "pretest": "standard",
+      "test": "node my-tests.js"
     }
   }
   ```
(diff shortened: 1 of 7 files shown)


charlike:
yep

dcousens:
any advantage?

charlike:
probably better support for windows, and not so long lines. but win already is using bash so..

LinusU:
I personally feel like the semantics of "pretest" is setting up something that is needed for the tests, e.g. seeding a database. In my mind this doesn't really change it for the better 🤔

ianwalter:
@LinusU I think that could be because that is how youve used it in the past? NPM does not tell you what the use case is. The docs simply tell you that pre and post are hooks that get run before and after their associated command. This is already a common pattern for a lot of people, see: http://www.marcusoft.net/2015/08/pre-and-post-hooks-for-npm-scripting.html

PR #970, opened 2017-08-09
fix error example 修复错误的示例

Airsola:
fix the error example on the rule 「eslint: brace-style」

### DIFF
--- a/docs/RULES-zhcn.md
+++ b/docs/RULES-zhcn.md
@@ -130,10 +130,12 @@
 
   ```js
   // ✗ avoid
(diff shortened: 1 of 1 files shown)


docs/RULES-zhcn.md
```diff
@@ -200,6 +202,7 @@
   ```js
   // ✗ avoid
   var value = 'hello world'
+  ```
```
LinusU: Was this really intended?

wayou: I dont' think so.

Issue #971, opened 2017-08-10
Add "--eject" command?


josephfrazier:
Standard is very useful, but sometimes I'd like to make small changes to the rules used. How do folks feel about the prospect of `standard --eject` converting the project to use `eslint` with `eslint-config-standard`, as described in https://github.com/standard/standard/tree/91e80a1feba31bb1c7b1ab57283c4e3cd93d56ab#i-disagree-with-rule-x-can-you-change-it?

This would be similar to the `npm run eject` command offered by `create-react-app`: https://github.com/facebookincubator/create-react-app/tree/10c1f577da211d65bcc278f94198ef75f00f0277#converting-to-a-custom-setup

timoxley:
You could probably prototype this as a separate package e.g. `standard-eject`

bcomnes:
Agreed a `standard-eject` would be super handy.  If you build that we can add it to the ecosystem section.

josephfrazier:
I just published an initial implementation: https://github.com/josephfrazier/standard-eject

I'll send a PR for adding it to the README.

josephfrazier:
Closing in favor of #973. Also, it looks like @flet's [standardx](https://github.com/standard/standardx) may fit the bill for many use cases.

PR #973, opened 2017-08-12
Mention `standard-eject` in README

josephfrazier:
See https://github.com/standard/standard/issues/971#issuecomment-321669543


Maksim2013:
yf[

Issue #966, opened 2017-08-08
Chinese linking is broken in readme.md.


Aladdin-ADD:

Flet:
This should be fixed now on GitHub and on standardjs.com! 👍

PR #975, opened 2017-08-21
Fix broken relative links in readme.md

chenxsan:
Links to translations are broken in readme.md. This pull request fix them by replacing relative path with absolute one.

Would fix issue https://github.com/standard/standard/issues/966


ematipico:
These links are used to generate the correct link on the website standardjs.com, I am not sure that we want to hardcode the github link

Flet:
FYI this should be fixed now :)

PR #1011, opened 2017-11-02
Add Eaze and Ctrl Alt Deseat to list of companies using standard

### LINKED ISSUE #1010: Add Eaze to the list of users
Someone I know who works there confirmed they use `standard`.

LinusU:
fixes #1010


LinusU:
Tests are failing because https://github.com/tunnckoCore/get-installed-path is using dangling commas...

I'll remove that repo from the list

LinusU:
Blocked on standard/standard-packages#16, we need to publish a new version of `standard-packages`...

Flet:
Does it make sense to add these to other langauge READMEs as part of this PR?

Also, can the two extra empty boxes be removed?
image

LinusU:
> Does it make sense to add these to other language READMEs as part of this PR?

Yeah, I'll fix!

> Also, can the two extra empty boxes be removed?

I don't think so, I can very possibly be wrong though, but this is how it looked before I added them:

screen shot 2017-11-02 at 10 03 13

LinusU:
@Flet I've updated the other languages 👌

PR #1017, opened 2017-11-13
Use as pretest in package.json

DiegoRBaquero:
They are pretty much the same, but semantically better, I feel.

### DIFF
--- a/docs/README-en.md
+++ b/docs/README-en.md
@@ -156,7 +156,8 @@ $ standard "src/util/**/*.js" "test/**/*.js"
       "standard": "*"
     },
     "scripts": {
-      "test": "standard && node my-tests.js"
+      "pretest": "standard",
+      "test": "node my-tests.js"
     }
   }
   ```


LinusU:
seems like the same thing as in #964

this was my personal stance then, and now:

> I personally feel like the semantics of "pretest" is setting up something that is needed for the tests, e.g. seeding a database. In my mind this doesn't really change it for the better 🤔 

> I haven't used the `pre*` or `post*` hooks much at all, but to me the linting is part of the test suite, not something that I run before it... just my 2¢

bcomnes:
I'm in the same boat as Linus, I've always considered linting as part of the set of tests I run on code.   I see the relationship, e.g. linting helps prevent bugs, and generally you want to run them before the test, but I also see the pre-test lifecycle as things needed to run before the tests that dont produce test results.  I guess this is the problem with semantics.

DiegoRBaquero:
I like how you see it.

PR #1020, opened 2017-11-25
tiny rewording for clarity

earthrid:
I believe it’s time to leave the “tl;dr” joke behind (in general) and
use a simple (standard… if not stylish) word that is easier to
translate and to understand for all: “summary”.


catdaddy23:
Kkk

feross:
@earthrid Thanks for the PR. I'm fine with this change, but it looks like this PR doesn't merge cleanly anymore. Sorry to bother you, but can you resend it?

Issue #1018, opened 2017-11-21
Git pre-commit hook doesn't work with BSD grep.


adriantrunzo:
The pre-commit hook was recently changed in commit bd5445c, but unfortunately the use of `grep -z` (after git-diff) doesn't seem valid for BSD grep. To note I am on macOS 10.13.1 and have verified the following on both the system bash (GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin17) and brew installed bash (GNU bash, version 4.4.12(1)-release (x86_64-apple-darwin16.3.0).

In this case the `-z` option is meant to handle null-terminated input lines created with `git-diff -z` (and later parsed by `xargs -0`). However, `grep -z` on macOS/BSD is defined as:

```
-Z, -z, --decompress
             Force grep to behave as zgrep.
```

That definition is obviously not want we want, and unfortunately seems to be subject to it's own issues, namely if you try to use the `-z` option you will see `grep: invalid option -- z`.

To make the pre-commit hook work in my setup I changed the `grep -z` to `grep -a --null`, which is not quite functionally equivalent (treating input as text vs null-terminated lines) but close enough to get things working. I believe the `-a` and `--null` options are the same across most grep versions.

Flet:
@ttencate thoughts on this?

ttencate:
Indeed `-z`/`--null-data` is not in the [POSIX standard](https://www.unix.com/man-page/posix/1p/grep/), my bad!

The `-Z`/`--null` flag is also nonstandard, but it is not what we need anyway: it _outputs_ a null byte after the _filename_, not after each _matched line_. At least, that's what it does on my system, Arch Linux with GNU grep 3.1:

>       -Z, --null
>              Output  a  zero  byte  (the  ASCII  NUL character)
>              instead of the character that normally  follows  a
>              file  name.   For example, grep -lZ outputs a zero
>              byte after each file name  instead  of  the  usual
>              newline.     [...]

Instead of nulls, we could use newlines (`\n`) as the delimiter, which is almost as safe as `\0`:

```
git diff --name-only --cached --relative | grep '\.jsx\?$' | xargs-r -d'\n' -t standard
```

But then we'd need the `-d`/`--delimiter` argument to `xargs`, which is _also_ [nonstandard](https://www.unix.com/man-page/posix/1p/xargs). In fact, `xargs -0` was never POSIX to begin with...

[...]

PR #1025, opened 2017-12-07
Update pre-commit hook to be more widely compatible.

adriantrunzo:
Contains changes from the discussion in issue #1018. Mainly use newlines instead of nulls and simplify read/echo handling in `xargs-r`.


docs/README-en.md
```diff
@@ -653,11 +653,11 @@ Funny you should ask!
 function xargs-r() {
   # Portable version of "xargs -r". The -r flag is a GNU extension that
   # prevents xargs from running if there are no input files.
-  if IFS= read -r -d '' path; then
```
azizol: May  mail

azizol: May icloud

PR #1027, opened 2017-12-11
update to eslint 4 in preparation for standard@11.0.0-beta.0

Flet:
- update to eslint 4
- update other dependencies
- also attempt to --fix repos with eslint 4 breaking changes

Hi folks!

This PR updated eslint and plugin/config dependencies to their latest versions. The idea is to release this as `standard@11.0.0-beta.0` since some existing eslint rules have gotten more strict. It will also be nice to update to eslint@4 to help folks who are running into various bugs/compatibility issues.

In order to get this out without a lot of hassle, we will not be including any new or modified rules in this major version. This does set the stage nicely for the next version though! :)

Also I included an update to `clone.js` that attempts to `--fix` repositories that have breaking changes related to eslint@4 updates. There are 23/235 (~10%) repositories that break with eslint@4, but they are `--fixed` successfully.

@feross @bcomnes @LinusU

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,45 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 11.0.0 - 2017-12-11
+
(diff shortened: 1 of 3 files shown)


followbl:
working great 👍

PR #1029, opened 2017-12-15
Fix typo on `no-label-var`

lacymorrow:
I'm pretty sure that the author meant to use an equal operator...


lacymorrow:
Good point, and it actually illustrates the label use. What do you think about simplifying the example even more?

```javascript
  score: while (true) {
    score -= 10
    if (score > 0) continue score
    break
  }

LinusU:
Sounds very good 👍

lacymorrow:
Looks like git doesn't like ticks in commit messages...
Updated, take a look and let me know what you think 👍

feross:
@lacymorrow Hey! Apologies that this never got merged. It looks like there are merge conflicts now, and rather than ask you to fix them up I tried to do so. However, I ran into issues and decided to just copy your changes into a new PR.

https://github.com/standard/standard/pull/1129

I hope this is okay. Thanks for the great work. Cheers!

PR #1043, opened 2018-01-14
Fix portuguese typo and some reference links

idlua:
It's just a small fix to the portuguese version of the README file.

There's an error in the word **_vírgulas-e-vírgula_**, where the correct is **_ponto-e-vírgula_**. Also, this fixes some reference links.


LinusU:
Thank you!

Issue #1078, opened 2018-02-24
npm ERR! peer dep missing: ajv@^6.0.0, required by ajv-keywords@3.1.0


kevzdev:
# EDIT: WORKAROUND IS TO INSTALL VERSION 10.0.3 `npm install standard@10.0.3 -g`

Problem occured when: `npm install standard -g`
Output:
```
C:\Users\Kevzdev\AppData\Roaming\npm\standard -> C:\Users\Kevzdev\AppData\Roaming\npm\node_modules\standard\bin\cmd.js
npm WARN ajv-keywords@3.1.0 requires a peer of ajv@^6.0.0 but none is installed. You must install peer dependencies yourself.

+ standard@11.0.0
added 224 packages in 4.381s

C:\Users\Kevzdev\Desktop\jsfolder>
```

`node -v` = `v9.5.0`
`npm -v` = `5.6.0`
Windows 10 Pro x64 v1709

It breaks NPM! When running `npm list -g --depth=0` output:
```
C:\Users\Kevzdev\AppData\Roaming\npm
`-- standard@11.0.0

npm ERR! peer dep missing: ajv@^6.0.0, required by ajv-keywords@3.1.0
```

kevzdev:
I've checked, `npm install standard -g`, `npm install standard@latest -g` and `npm install standard@11.0.0 -g ` all have this issue. Seems to be an issue with `standard@11.0.0`

**workaround is to install version 10.0.3: `npm install standard@10.0.3 -g`**

kevzdev:
Issue persists on both Windows 10 Pro x64 (version 1709) and macOS High Sierra (version 10.13.3) with node versions `8.9.4` (LTS) and `9.5.0` (current) and npm version `5.6.0` (current).

kevzdev:
It's probarely an issue in a dependency: https://www.diffchecker.com/mMdOuh7l (compared 10.0.3 to 11.0.0)

kevzdev:
I've narrowed down the issue to the dependency `"eslint": "~4.18.0"`.
This is a working version `"eslint": "~3.19.0"` however `"eslint-config-standard": "11.0.0"` and `"eslint-config-standard-jsx": "5.0.0"` depends on `"eslint": "~4.18.0"` so we need to downgrade them aswell to `"eslint-config-standard": "10.2.1"` and `"eslint-config-standard-jsx": "4.0.2"` since they depend on `"eslint": "~3.19.0"`.

kevzdev:
[...]

PR #1079, opened 2018-02-24
Update package.json

kevzdev:
Fixes https://github.com/standard/standard/issues/1078

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "standard",
   "description": "JavaScript Standard Style",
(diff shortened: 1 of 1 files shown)


ematipico:
We cannot downgrade to eslint 3.*! The aim of `standard@4.0.0` was to upgrade eslint to 4! We have to find another solution. Could you provide a repo to reproduce your issue?

feross:
See https://github.com/standard/standard/issues/1078#issuecomment-369092782

PR #1090, opened 2018-03-12
Swap README.md and RULES.md symlinks

### LINKED ISSUE #1074: Main readme is not updating on npmjs.com
The readme on https://www.npmjs.com/package/standard is quite out of date.

According to @zkat on twitter https://twitter.com/maybekatz/status/966085281802629121, npm doesn't publish symlinks. So the website readme just stopped updating.

We should consider moving the English readme back to the root. We'll need to update the website build scripts to account for this change.

watson:
@feross I hope it's ok I took a stab at this low hanging fruit. I decided to move the RULES.md as well so the symlinks go the same way. Thought it was easier to reason about.

Fixes #1074


feross:
Thanks for the PR @watson! I think this has the potential to affect the website (https://github.com/standard/standard-www).

@Flet - does this look good to you?

Flet:
Wow I totally missed this issue :)

OK I pulled this PR down and tried it in a local build of`standard-www` and it worked just fine 👍

PR #1091, opened 2018-03-14
Update Typeform's logo

mackermans:
We have a new logo as a result of our rebranding, so it would be great to see it updated here too.


LinusU:
official statement: https://www.typeform.com/blog/inside-story/rebranding/

PR #1092, opened 2018-03-14
Replace Opbeat with Elastic logo

watson:
Opbeat have been acquired by Elastic and the Opbeat brand is shutting down in a few months.

Proof of use: https://github.com/elastic/apm-agent-nodejs/blob/fda015a8a5fff38839c1a0e53e4b2ee3fcb161cc/package.json#L8


LinusU:
official statement: https://opbeat.com/blog/posts/welcome-opbeat-to-the-elastic-family/

Issue #1099, opened 2018-03-26
False positive `no-undef` in TypeScript with interface


kinyat:
see: https://github.com/eslint/typescript-eslint-parser/issues/416

The problem with using StandardJS is that we do not have control on the rules thus we cannot turn off `no-undef`.

I hope we can add some config for TypeScript to turn this rule off.

Or allow user to turn off some of the rules.

ColtonProvias:
It should be noted that abstract classes and class members inherited from abstract classes also throw this error.

lukasoppermann:
Having the same issue, @feross is there a solution for this yet? Isn't there a way to specify configs for this somehow? Due to this issue I can't use standard anymore with typescript. Which sucks because standard is really awesome.

Please help.

kevrat:
+1
![2018-07-03 16-42-20](https://user-images.githubusercontent.com/9952048/42223386-4d92a3c4-7ee0-11e8-96ea-9d94074096f0.png)

kevrat:
Also for no-use-before-define
![2018-07-03 16-45-55](https://user-images.githubusercontent.com/9952048/42223606-d0bfcbbe-7ee0-11e8-88fc-50a2c5db4e0e.png)

kinyat:
JYI: https://github.com/standard/standard/pull/1101

I have created a PR to turn this off for typescript but it hasn't been reviewed/approved and it is stale

mbret:
Any news on that ? The issue still exist and makes standardjs unusable with typescript.

feross:
Let's track this issue here: https://github.com/standard/standard/issues/1283

Issue #1007, opened 2017-10-27
Adopt eslint-config-standard-with-typescript


mightyiam:
I've been working on [eslint-config-standard-with-typescript](https://www.npmjs.com/package/eslint-config-standard-with-typescript) alone. It would be nice to have it adopted by the standard org and have some cooperation on it.

mightyiam:
👅

LinusU:
Hey, sorry that no one has responded to you. I would be happy to help you transition the package and help maintain it as well 🙌

mightyiam:
What do we do?

LinusU:
Saw that you invited me, I'll try and move it here 🙌

LinusU:
ah, I cannot do that, but I think that you can initiate a transfer, and then I can accept it on our side

mightyiam:
"You don’t have the permission to create repositories on standard" when trying to transfer it to the standard organization.

mightyiam:
Try adding me to the standard organization?

mightyiam:
I'm totally up for this to be adopted by @standard.

PascalAOMS:
Maybe it progresses a bit faster if I leave a comment here. /shrug

mightyiam:
@PascalAOMS the comment is in response to the stale bot adding the `stale` label.

I am still actively maintaining this package. And I've just updated it with all the breaking changes from upstream https://github.com/typescript-eslint/typescript-eslint/.

LinusU:
Sorry for the delay here, I think that @feross or someone else that is `owner` on the standard reop will need to fix this 😬 unfortunately I don't have the permissions to fix this

mightyiam:
:trumpet:

despian:
+1 would love to see this integrated into standard

bjufre:
Any news on this?

swansontec:
I tried using `eslint-config-standard-with-typescript`, but found that the rules are much stricter than I would have expected. For instance, it requires every function to declare its return type, even though Typescript is perfectly capable of inferring this. It also enforces the `number[]` array style over `Array<number>`, even thought the later is perfectly valid.

I understand that `eslint-config-standard-with-typescript` is derived from the `@typescript-eslint/eslint-plugin` recommended preset. Unfortunately, that preset has a rather aggressive "everything but the kitchen sink" approach to enabling linting rules.

[...]

PR #1101, opened 2018-03-28
Exclude typescript from no-undef checking

kinyat:
Fix for https://github.com/standard/standard/issues/1099

It's a [known issue](https://github.com/eslint/typescript-eslint-parser#known-issues) in `typescript-eslint-parser`

The solution is based on [here](https://github.com/eslint/typescript-eslint-parser/issues/416#issuecomment-363115171)
(Note: `**.*.ts` and `**.*.tsx` do not seems to work but `*.ts` and `*.tsx` work)

`ts` means typescript and `tsx` means typescript with react

Tested with my project and it successfully exclude no-undef check for typescript only

Before

image

After

image

### DIFF
--- a/eslintrc.json
+++ b/eslintrc.json
@@ -1,3 +1,9 @@
 {
-  "extends": ["standard", "standard-jsx"]
+  "extends": ["standard", "standard-jsx"],
+  "overrides": {
+    "files": ["*.ts", "*.tsx"],
+    "rules": {
+      "no-undef": "off"
+    }
+  }
 }


SinoBoeckmann:
Any plans to accept this pr? :)

LinusU:
I think that this is fixed upstream, we should just have to enable this rule: https://github.com/nzakas/eslint-plugin-typescript/blob/master/docs/rules/no-unused-vars.md

dcousens:
LGTM,  this would resolve the existing typescript problems when using `interface`s etc,  but,  I don't think it is our problem.

@LinusU are you saying that upstream have fixed it without any work on our end?  Or do we need to enable some rule?

LinusU:
> Or do we need to enable some rule?

It seems like we need to enable the `typescript/no-unused-vars` rule, which should "Prevent TypeScript-specific constructs from being erroneously flagged as unused".

dcousens:
Related https://github.com/standard/standard/issues/1099

dcousens:
@feross  no doubt,  this is supremely frustrating for TS devs

delvedor:
This would be super useful indeed! :)

cdaringe:
There is a new set of ts plugins in town, and they do fix the issue when using pure eslint.  However, the new plugins don't register correctly with standard.  I've recorded my current state in the below issue if someone wants to take a peek

https://github.com/standard/standard/issues/1252

Eomm:
What is blocking this pr?

It would be awesome don't be forced to migrate to eslint only for this rule that a lot of ts users are encountering. we love standard thanks to "zero config" feature

despian:
There are a number of other compatibility issues with the standard js rules and typescript. There was a suggestion in #1007 to merge rules from [eslint-config-standard-with-typescript](https://github.com/mightyiam/eslint-config-standard-with-typescript). That seems like a better solution to me.

feross:
Let's discuss a more all-encompassing support for TypeScript in this issue: https://github.com/standard/standard/issues/1007

PR #1111, opened 2018-04-16
clarified the documantation for vim with ale

andresattler:
ALE from default is not happy without a eslint config file. 
So for using standard as intended this few lines of configuration can be added to fix this.


feross:
Going to assume that this is correct and merge! 🤞

If any vim users disagree, please chime in.

PR #1128, opened 2018-05-10
Add an issue template

feross:
What do folks think about this?


feross:
No response - merging

PR #1141, opened 2018-05-28
Replace RawGit URLs

GeorgeBgk:
RawGit is dead, there's a similar service. RIP RawGit.


GeorgeBgk:
Anyone here?

GeorgeBgk:
Oops.. It might be my ISP blocking rawgit's IP addresses in an attempt to ban Telegram. I thought that rawgit was dead 😅

PR #1145, opened 2018-06-04
docs: vim ale standard fixer

theo4u:
As ale has been updated to handle file fix on save for standard https://github.com/w0rp/ale/pull/781

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,11 +286,12 @@ Install **[ale][vim-1]**. And add these lines to your `.vimrc` file.
 let g:ale_linters = {
 \   'javascript': ['standard'],
 \}
+let g:ale_fixers = {'javascript': ['standard']}
 ```
 
-This sets standard as your only linter for javascript files and so prevents conflicts with eslint.
+This sets standard as your only linter and fixer(on save) for javascript files and so prevents conflicts with eslint.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -286,11 +286,12 @@ Install **[ale][vim-1]**. And add these lines to your `.vimrc` file.
...
 \   'javascript': ['standard'],
 \}
+let g:ale_fixers = {'javascript': ['standard']}
 ```
 
-This sets standard as your only linter for javascript files and so prevents conflicts with eslint.
+This sets standard as your only linter and fixer(on save) for javascript files and so prevents conflicts with eslint.
 
-For automatic formatting on save, add these lines to `.vimrc`:
+For automatic formatting on save, add these lines to `.vimrc` if you are not using **[ale][vim-1]**:
```
rstacruz: I'd remove these altogether and just recommend using `ale`. The old method described here is pretty unusable anyway, since it locks up Vim for a couple of seconds every save!

rstacruz: Instead consider changing that to something like

> For automatic formatting on save, use ale's `g:ale_lint_on_save` setting:
>
> ```vim
> let g:ale_lint_on_save=1
> ```

feross:
@rstacruz Looks like the feedback was addressed, so I'll merge this!

PR #1162, opened 2018-07-10
translate README in French

charliegerard:


feross:
Thanks, this is merged and released on the website now! ✨

If you're feeling up for it, there's also a `RULES.md` that can also be translated. You can find those files in the `docs/` folder. For example, here is the English version: https://github.com/standard/standard/blob/master/RULES.md

charliegerard:
Thanks @feross !! :)

I started translating the RULES.md and am about half way through, so will send a pull request soon!

feross:
That's awesome, Charlie!

Issue #1166, opened 2018-07-24
babel-eslint undefined


talmobi:
**What version of standard?**

11.0.1

**What operating system, Node.js, and npm version?**

`uname -a` `Darwin Kernel Version 17.6.0`
`node -v` `v8.11.2`
`npm -v` `6.1.0`

**What did you expect to happen?**

Globally installed standard to work out of the box `npm install -g standard`

**What actually happened?**

An error is thrown: `Cannot find module 'babel-eslint' (eslint/undefined)`

Related issue: `https://github.com/standard/standard/issues/85`

According to @feross both `standard` and `babel-eslint` needs to be installed (ref: `https://github.com/standard/standard/issues/85#issuecomment-391155461`).

Then why is `babel-eslint` only set as a `devDependency` in the `package.json`?

 Seems like it should be a normal `dependency`

Installing `babel-eslint` globally as well resolves the issue: `npm install -g babel-eslint`

PR: `https://github.com/standard/standard/pull/1167`

talmobi:
Issue resolved at https://github.com/standard/standard/pull/1167

Closing.

Issue #85, opened 2015-03-25
Cannot find module 'babel-eslint' (eslint/undefined) 


mattdesl:
Using this as a local dependency has the following error:

```
Error: Use JavaScript Standard Style (https://github.com/feross/standard)
  cmd.js:0:0: Cannot find module 'babel-eslint'
```

Installing it globally and running that works. 

Getting the same error in the SublimeLint plugin. 

Versions:
```
standard@3.2.1
└── eslint@0.17.1 
```

mattdesl:
Doing `npm install babel-eslint --save-dev` seems to fix the issue as well.

Flet:
Really strange, I don't believe `babel-eslint` is used anywhere in this repo or any branches.

feross:
Weird. Can you try again with the latest version, 3.3.0?

mattdesl:
works now :+1:

baygeldin:
Hi! I've just faced with this problem and it turned out that `standard` was installed globally and `babel-eslint` was installed into `nvm` internal folder. Be careful when using `nvm` :)

sethbergman:
Try this setting:

![](https://cloud.githubusercontent.com/assets/427137/14826214/b85fabf8-0b91-11e6-9786-8a6ea0d1affe.png)

You don't need to add `babel-eslint` as a dependency. Just check this setting in Atom.

otajor:
Just had the same error.

I had:
- `standard` installed globally but not locally as a dependency in my project
- `babel-eslint` installed locally but not globally

Once I added standard as a local dependency as well, it was fine.

chevdor:
FYI:
Added with `npm i -D babel-eslint` and it still failed.
Installed globally `npm i -g babel-eslint` and it worked.

feross:
Yep, both `standard` and `babel-eslint` need to be installed at the same level, both local or both global.

chevdor:
Even both local does not work.

feross:
@chevdor Please open a new issue with all the details of your situation.

PR #1167, opened 2018-07-24
moved `babel-eslint` to dependencies

talmobi:
`babel-eslint` is now an explicit dependency as opposed only a devDependency like it used to be. Without this change `npm install -g standard` will not work out of the box without also installing `npm install -g babel-eslint`

github issue ref: https://github.com/standard/standard/issues/1166

**steps to reproduce error**

`npm uninstall -g standard babel-eslint`
`npm install -g standard`
`standard`
`...index.js:0:0: Cannot find module 'babel-eslint'`

**steps to try out fix**

`npm uninstall -g standard babel-eslint`
`npm install -g talmobi/standard#master`
`standard`
`.../Users/mollie/work/TICKETPLAT-fe-admin/src/components/templates/AncillaryTickets/index.js:95:16: There should be no spaces inside this paren.`

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,10 +20,10 @@
     "eslint-plugin-promise": "~3.8.0",
     "eslint-plugin-react": "~7.7.0",
     "eslint-plugin-standard": "~3.1.0",
+    "babel-eslint": "^8.0.3",
     "standard-engine": "~9.0.0"
   },
   "devDependencies": {
-    "babel-eslint": "^8.0.3",
     "cross-spawn": "^6.0.3",
     "eslint-index": "^1.3.0",
     "minimist": "^1.2.0",


LinusU:
Hmm, the usage of babel-eslint is optional though, right?

talmobi:
@LinusU According to https://github.com/standard/standard/issues/85#issuecomment-391155461 both needs to be installed?

And if you install `standard` by itself it doesn't work.

If it's optional, then why is it necessary? Doesn't that seem a bit like the opposite of optional? Maybe I'm missing something?

LinusU:
babel-eslint is only used if the `"parser"` option is set to `"eslint"`, e.g. by setting `"standard": { "parser": "babel-eslint" }` in the `package.json`.

The default is to use the builtin parser from eslint.

talmobi:
@LinusU Just tested on a fresh Ubuntu droplet and you are indeed correct.  The projects I tested against had the `"standard": { "parser": "babel-eslint" }`  set in `package.json`.

Perhaps standard could check the local `node_modules` belonging to the `package.json` it is parsing? Or perhaps a few tips added to the missing dependency error?

Anyway, this matter is resolved -- thank you for your time @LinusU

LinusU:
The problem is probably that you were running the global standard with the local config.

- Make sure that the project has both `standard` and `babel-eslint` in the `devDependencies` section of the `package.json`.
- Make sure that you are running the local installation of `standard` (e.g. `npm test`, `npm run lint`, ...)

talmobi:
Indeed. The issue in a sense was that I wanted to specifically run the global `standard` on the project without consulting the projects `package.json` and its `babel-eslint` dependency. Perhaps using standard globally should not interpret project specific `package.json`? Not sure..

Anyway, thanks so much for the detailed help and responses~

talmobi:
Or perhaps a way to turn off the `opts.usePackageJson` https://github.com/standard/standard-engine/blob/master/index.js#L96 with a CLI argument?

PR #1172, opened 2018-08-03
docs: add badgen standard style badge

clarkdo:
Add badgen as an alternative chocie of badge, it's faster and more stable.


feross:
Thanks for the PR, but this looks exactly the same as what we already have.

PR #1190, opened 2018-08-30
Silently pass in node 4

Gerhut:
Since [ESLint 5 no longer support node 4](https://eslint.org/docs/user-guide/migrating-to-5.0.0#-nodejs-4-is-no-longer-supported), standard should silently pass in node 4.

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
-if (process.version.match(/v(\d+)\./)[1] < 4) {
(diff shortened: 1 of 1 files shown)


Gerhut:
BTW, Node.js [encourages](https://nodejs.org/en/blog/release/v8.0.0/#long-term-support) using Node.js 8 without "v" in front of the version number, should we keep v of v6 in console logs?

feross:
Released as 12.0.1.

PR #1193, opened 2018-09-04
Translate RULES.md in french

charliegerard:
Translated the rules in french and added the link in other files


feross:
Thanks! This is live on https://standardjs.com/rules-fr.html now!

PR #1195, opened 2018-09-07
No 'ok' sign at 442 line

dangen-effy:
No 'ok' comment exists at 442 line.

I modified it.


dangen-effy:
Hi. thanks for review. But not merged yet? :o

PR #1198, opened 2018-09-11
es6 style

jimmywarting:
converted using

```
lebab --replace <FILE> --transform arrow,let,includes,template,default-param,obj-shorthand,obj-method,arg-spread,arg-rest,for-each,class
standard --fix <File>
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,4 @@
-var Linter = require('standard-engine').linter
-var opts = require('./options')
+const Linter = require('standard-engine').linter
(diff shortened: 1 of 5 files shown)


feross:
@jimmywarting Thanks for this PR! Sorry I didn't see it in time. Now it looks like it has merge conflicts, so I'm going to just send a separate PR to do this. Thanks again!

feross:
See https://github.com/standard/standard/pull/1353

PR #1210, opened 2018-10-09
Bump eslint to 5.6.0 (compat with react-scripts)

jtrussell:
This minor version bump of eslint is for compatability with react-scripts@2.x which
insists on this version of eslint on startup.

This change will allow folks to continue using standard in conjunction with new react-scripts
releases without issue.


LinusU:
I think this should be good to merge 👍

PR #1224, opened 2018-11-10
Add a disambiguating "standardjs" bin alias

searls:
Because there are now "standard"-branded opinionated linters being 
written for other languages, it seems smart to give people who use both
a way to disambiguate between the two by running `standardrb`, 
`standardjs`, etc. explicitly.

Someday, it'd be really neat if a meta `standard` bin could roll-up 
linting for multiple languages in a single project in a single
invocation, but for now I think this is a decent way to remediate the
risk of people globally installing multiple standard packages and 
invoking each from the user PATH.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,10 @@
     "name": "Feross Aboukhadijeh",
     "url": "https://feross.org/"
   },
-  "bin": "./bin/cmd.js",
+  "bin": {
+    "standard": "./bin/cmd.js",
+    "standardjs": "./bin/cmd.js"
+  },
   "bugs": {
     "url": "https://github.com/standard/standard/issues"
   },


tunnckoCore:
Agree. Make sense.

feross:
@searls I believe when we discussed this over DM we agreed to not merge this, right? Lmk if something has changed.

searls:
@feross The immediate urgency is definitely gone, because I renamed the bin `standardrb` to avoid the conflict.

However, I'd still really like to merge this, even though the going concern is not there. Because Ruby `standard` has the gem name `standard`, a lot of people will get easily confused on projects where `standard` points to npm `standard` but `standard`-the-gem points to `standardrb`. 

I think it'd make it just a little bit easier if `standardjs` was a common alias, so that when I'm writing something like a CI script I can be explicit in saying `standardjs` and `standardrb` instead of the reader needing to remember that your `standard` came first

feross:
> a lot of people will get easily confused on projects where standard points to npm standard but standard-the-gem points to standardrb.

Curious if you have seen this in practice, or is this primarily a theoretical concern at this point?

searls:
@feross it's tripped me up once or twice, but that's because I'll have both installed and then forget I renamed the Ruby bin (🤦‍♂). But this has been cited to me a few times as a barrier/friction when Ruby+JS teams try to adopt both; I guess the expectation of lib name uniqueness transcends language 🤷‍♂

PR #1240, opened 2019-01-16
Add pointers for those who disagree with standard

brodybits:
to semistandard & standardx

Note that these are already listed at the end of README.md. It might be a little smarter to just make a single list of alternatives with an anchor link.


brodybits:
Build seems to be broken by standard/standard-packages#25, hope someone can merge that one.

PR #1246, opened 2019-02-06
Show npx usage example

Ethan-Arrowood:
I'd like to propose an npx usage example. Let me know if this is the right place for it or if I should add more context to the example.


LinusU:
Personally not a fan of putting `npx` in the `"scripts"` section. One of the problem is that it wouldn't pin the version. 🤔

DiegoRBaquero:
`npx` is not needed in the scripts section if you have the modules in your dev dependencies installed

PR #1261, opened 2019-03-25
correct grammatical errors

MareenaKunjachan:


LinusU:
I'm not sure that those commas were grammatically incorrect? 🤔

Would you mind elaborating on why you think they are?

feross:
Thanks for this PR, but I think the sentences are fine the way they are. Please feel free to keep sending other PRs if you'd like.

PR #1269, opened 2019-04-02
Node.js API (Removed headings on pre-formatted text and restructuring)

co16353sidak:
1. Removed headings from pre-formatted syntax
2. Attempted to restructure some content in order to enhance readability.
3. as discussed on standard/standard-www#28

Fixes standard/standard-www#28

### DIFF
--- a/README.md
+++ b/README.md
@@ -714,9 +714,12 @@ There's also [standard-tap](https://www.npmjs.com/package/standard-tap),
 
 Yes!
 
-### `standard.lintText(text, [opts], callback)`
+```js
+standard.lintText(text, [opts], callback)
+```
+Lints the provided source `text`. 
 
-Lint the provided source `text`. An `opts` object may be provided:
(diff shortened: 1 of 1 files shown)


co16353sidak:
Hey @LinusU waiting on some feedback on this, want to resolve this so that i could work on another PR here
please resolve lest it becomes stale

LinusU:
Sorry for the delay on this!

I know that I initially said that it was a bit weird that they were headings, but after examining it again I actually think it's quite nice. I think it's quite common to write api documentation in markdown that way, it makes it more structured around the different functions instead of reading as a big block of text.

This also seems to be how Node.js itself does it: https://github.com/nodejs/node/blob/master/doc/api/os.md

Would love it if @feross or @Flet could give some input, also open to more discussion from anyone ☺️

feross:
I appreciate the effort but this doesn't add anything.

PR #1288, opened 2019-06-03
Translate rules into Simplified Chinese

Amabel:
Hi, I found there was a sentence not being translated into Chinese. Could you please take a look at this?


feross:
Thank you!

PR #1308, opened 2019-07-05
Update Japanese documents to follow original documents

munierujp:
I've updated Japanese documents to follow original documents (from da127f2 to 40d1d5e)


feross:
@munierujp Thank you for the excellent work! I have added you as a collaborator to the @standard organization!

munierujp:
@feross Thank you for inviting 😄 I can mainly maintain Japanese documents 🇯🇵

feross:
@munierujp That would be excellent.

PR #1346, opened 2019-08-02
Fix typo and following conventions

dangen-effy:
javaScrpt to JavaScript
typescript to TypeScript

- Following case convention (referencing en doc)
- Correct typo


feross:
Thanks @dangen-effy!

PR #1353, opened 2019-08-12
Use modern ES style

feross:
Supersedes https://github.com/standard/standard/pull/1198 by @jimmywarting. Thanks for the handy lebab command!


feross:
Merging without review because I want to try to do a standard v14 beta release soon

PR #1387, opened 2019-08-24
Update README.md

RealRichardParker:
Now with better install instructions.


LinusU:
These install instructions seems better suited for the other package's readme, since that is what it would be installing...

PR #1393, opened 2019-08-28
Update package.json

EasyGrease:
Remove "funding" adware.

"funding" is a "paused" experiment and this commit removes it as a dependency from standard/standard, least ads come back in build logs unexpectedly.  By leaving this dependency in place it allows feross/funding to turn ads back on in standard/standard without any changes in standard/standard, this commit prevents that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,6 @@
     "eslint-plugin-promise": "~4.2.1",
     "eslint-plugin-react": "~7.14.2",
     "eslint-plugin-standard": "~4.0.0",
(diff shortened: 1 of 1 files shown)


mightyiam:
Hi, @EasyGrease. Thank you for using standard and for your pull request. As you probably know, the ad raised concerns and it seems that those have been voiced in the issue tracker already. Please allow me to view this PR as more feedback. I'm sure it had reached its destination successfully.

PR #1396, opened 2019-09-03
Add Grab to users list

LitoMore:
Hi Standard!

We are Geo Tools team in Grab. We are using JavaScript Standard Style for linting our front-end projects. That's awesome! Thank you!

I also updated the same section in other languages for you!


feross:
Very nice! And thank you for updating the logos section in all the language files. Cheers!

PR #1398, opened 2019-09-04
changelog: Run 'npx vweevers/hallmark'

feross:
This massively improves the changelog. Caught lots of inconsistencies. Love it! 

All these changes were automatically fixed by `hallmark` using the command from https://github.com/standard/standard/commit/a7762a2ac0f8c5e43000ccc33304613cfb5f6569#commitcomment-34902423

There's still lots of other issues it is reporting that are not fixed by this PR, but I wanted to get these changes in regardless, since they're nice.

cc @vweevers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,13 @@
 # Changelog
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [Unreleased][unreleased]
 
-## 14.1.0 - 2019-08-28
+## [14.1.0] - 2019-08-28
 
 - Support ES 2020, the latest version of the ECMAScript specification, which includes support for Dynamic Imports and `BigInt`. [#1378](https://github.com/standard/standard/issues/1378)
 - Security: Upgrade `eslint` to `6.2.2` to fix security issue ([blog post](https://eslint.org/blog/2019/08/eslint-v6.2.1-released)) ([security advisory](https://github.com/mysticatea/eslint-utils/security/advisories/GHSA-3gx7-xhv7-5mx3))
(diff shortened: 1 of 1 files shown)


feross:
As for the remaining issues, they seem to be broadly two main issues:

- Found reference to undefined definition
  - Caused by the percentage breakage that we include next to new rules. They're in square brackets. We could change to parens to solve the problem.

- Group heading must be one of Changed, Added, Deprecated, Removed, Fixed, Security
  - This is harder since we tend to distinguish between 'New features' and 'New rules' in our changelogs. I suppose we could use Added for both and then have subheadings for Added Rules and Added Features?

CHANGELOG.md
```diff
@@ -644,170 +646,351 @@ In `package.json`, use the "standard" property:
...
+
+[6.0.6]: https://github.com/standard/standard/compare/v6.0.5...v6.0.6
+
+[6.0.5]: https://github.com/standard/standard/compare/v6.0.4...v6.0.5
+
+[6.0.4]: https://github.com/standard/standard/compare/v6.0.3...v6.0.4
+
+[6.0.3]: https://github.com/standard/standard/compare/v6.0.2...v6.0.3
+
+[6.0.2]: https://github.com/standard/standard/compare/v6.0.1...v6.0.2
```
ungoldman: @vweevers only minor thing I see here is the formatter's producing a lot of newlines in between each link reference, might be nice to compact this area a bit. otherwise i really like it, thanks for making this happen!

feross: Yeah, agreed.

vweevers: Agree. It's a `remark` default, I haven't looked into it yet.

vweevers:
> We could change to parens to solve the problem.

That makes sense. The changelog also states "based on test suite in parens".

> we tend to distinguish between 'New features' and 'New rules' in our changelogs. I suppose we could use Added for both and then have subheadings for Added Rules and Added Features?

Though subheadings would work, because I'm currently not linting the content below the headings, it would make it harder to lint that content in the future.

Alternatively, you could prefix changes with e.g. "- Add rule:".

vweevers:
Couple of things to be aware of:

- For adding a new release, you can add a release header (e.g. `## 1.0.0 - 2019-09-05`), run `hallmark --fix` to populate the release with a commit log, then tweak that by hand.
- Known issue: `hallmark --fix` adds GitHub links to text like `#23`, but if `hallmark` itself generated markdown content (like a release) you will need to run `hallmark --fix` twice.
- If you want to run `hallmark` on your README as well, you will need these [Package options](https://github.com/vweevers/hallmark#package-options): `{ paddedTable: false, toc: false }`

feross:
@vweevers I'm going to just merge this PR since I think it improves things. We can try adding the `hallmark` package itself, as well as linting the README, in a future PR.

PR #1404, opened 2019-09-06
Format AUTHORS.md with hallmark

vweevers:
**What is the purpose of this pull request?**

[x] Documentation update

**What changes did you make?**

- Changed `tools/update-authors.sh` to wrap email addresses in angle brackets (`<>`) which ensures `remark` recognizes them as links (rather than literal URLs, which is a GFM feature) which in turn prevents the bug https://github.com/remarkjs/remark-github/issues/18.
- Added `hallmark` (only on `AUTHORS.md` to start) to format email addresses as `mailto:` links.


feross:
Is there a reason that `hallmark` doesn't just leave the the emails formatted like `<user@example.com>`? It seems cleaner to leave them that way if they're recognized as emails by GitHub and other tooling.

vweevers:
The idea behind it is to render the markdown the same as GitHub would (in places that aren't GitHub).

feross:
@vweevers Understood.

PR #1405, opened 2019-09-08
Add no-unexpected-start

### LINKED ISSUE #1402: Disallow semicolons before statement continuation characters
**What version of this package are you using?** 14.1.0

**What problem do you want to solve?** I'd like to continue the semicolon purge by disallowing semicolons as a workaround for lines that start with ` [`, `(`, `` ` ``, `+`, `*`, `/`, `-`, `,`, or `.`. In [other words](https://standardjs.com/rules.html#semicolons):

> Note: If you're often writing code like this, you may be trying to be too clever.
>
> Clever short-hands are discouraged, in favor of clear and readable expressions, whenever possible.

**What do you think is the correct solution to this problem??**

```json
{ "beforeStatementContinuationChars": "never" }
```

Before:

```javascript
;(function () {
  window.alert('ok')
}())
 
;[1, 2, 3].forEach(bar) 

;`hello`.indexOf('o')
```

After:

```javascript

{
  // ES6 curly jail
  // All the benefits of IIFE without the mess
  window.alert('ok')
}

var nums = [1, 2, 3]
nums.forEach(bar)

var word = `hello`
word.indexOf('o')
```

**Are you willing to submit a pull request to implement this change?** Absolutely.

christianbundy:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Resolves #1402

Nothing in particular, currently just a draft to work on the documentation changes suggested by @mightyiam. In the future this PR should *actually* add the relevant rule change.

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -1323,25 +1323,13 @@ your code.
 
   This is the only gotcha with omitting semicolons, and `standard` protects you from this potential issue.
 
-  (The full list is: `[`, `(`, `` ` ``, `+`, `*`, `/`, `-`, `,`, `.`, but most of these will never appear at the start of a line in real code.)
-
   eslint: [`no-unexpected-multiline`](http://eslint.org/docs/rules/no-unexpected-multiline)
+  eslint: [`no-unexpected-start`](http://eslint.org/docs/rules/no-unexpected-start)
 
   ```js
   // ✓ ok
-  ;(function () {
-    window.alert('ok')
-  }())
-
-  // ✗ avoid
-  (function () {
-    window.alert('ok')
-  }())
-  ```
-
-  ```js
-  // ✓ ok
-  ;[1, 2, 3].forEach(bar)
+  var nums = [1, 2, 3]
+  nums.forEach(bar)
 
   // ✗ avoid
   [1, 2, 3].forEach(bar)
@@ -1349,30 +1337,15 @@ your code.
 
   ```js
   // ✓ ok
-  ;`hello`.indexOf('o')
+  {
+    window.alert('ok')
+  }
 
   // ✗ avoid
-  `hello`.indexOf('o')
+  (function () {
+    window.alert('ok')
+  }())
   ```
-
-  Note: If you're often writing code like this, you may be trying to be too clever.
-
-  Clever short-hands are discouraged, in favor of clear and readable expressions, whenever
-  possible.
-
-  Instead of this:
-
-  ```js
-  ;[1, 2, 3].forEach(bar)
-  ```
-
-  This is strongly preferred:
-
-  ```js
-  var nums = [1, 2, 3]
-  nums.forEach(bar)
-  ```
-
 
 ## Helpful reading
 


RULES.md
```diff
@@ -1323,57 +1323,30 @@ your code.
...
-  ```
-
-  This is strongly preferred:
-
-  ```js
-  var nums = [1, 2, 3]
-  nums.forEach(bar)
+  (function () {
+    window.alert('ok')
+  }())
```
christianbundy: @mightyiam 

> How about we keep an example of an IIFE and then show the block alternative?

Does this look alright to you?

RULES.md
```diff
@@ -1323,57 +1323,30 @@ your code.
...
-  (The full list is: `[`, `(`, `` ` ``, `+`, `*`, `/`, `-`, `,`, `.`, but most of these will never appear at the start of a line in real code.)
-
   eslint: [`no-unexpected-multiline`](http://eslint.org/docs/rules/no-unexpected-multiline)
+  eslint: [`no-unexpected-start`](http://eslint.org/docs/rules/no-unexpected-start)
 
   ```js
   // ✓ ok
-  ;(function () {
-    window.alert('ok')
-  }())
```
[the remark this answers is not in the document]
christianbundy: :arrow_right: https://github.com/standard/standard/pull/1405/files#r322404768

christianbundy:
> What about semicolons that separate between statements that are on the same line? [..] Would we like to disallow them?

Personally, I'd love to disallow them, but I'm not sure whether it's worthwhile for StandardJS to disallow them. It's also currently [not impossible](https://github.com/eslint/eslint/issues/12228), since semi's `"never"` option actually means "never, except when it protects against an ASI hazard", although it's possible that will be changed in the future.

> And what about the semicolons that are between the parts of a for statement?

If we want StandardJS to be completely free of semicolons we could disallow for-loops in favor of while-loops, but I'm not sure whether that's really worth fighting for. I'm in favor of using a subset of language features, and *maybe* it would be nice to get rid of semicolons in general, but think it should probably be discussed more (and likely added in a different PR).

RULES.md
mightyiam: Kind of. When I read this code snippet, I don't see a good reason for the `window.alert('ok')` to be in an IIFE. Is there a piece of code that makes it obvious why it should be in an IIFE?

mightyiam:
We need more buy-in and also the rule implementation.

christianbundy:
I'm not sure whether ESLint core is open to having this rule added. Should this be added to the StandardJS ESLint plugin or should I make my own plugin with the rule? I don't feel strongly one way or the other.

mightyiam:
@christianbundy I value your contributions. I would expect that ESLint core are willing to accept this rule in a form that they see appropriate, whether it be a new rule or an addition to an existing one. Could you please link to the relevant ESLint conversation, here?

christianbundy:
It's just this discussion: https://github.com/eslint/eslint/issues/12228, I've summarized my understanding [here](https://github.com/eslint/eslint/issues/12228#issuecomment-530173818).

There's some contention over the fact that `{ "semi": ["error", "never"]` }` allows semicolons, and whether that's a bug, but it doesn't sound like it will be fixed. I can make a pull request for the `{ "semi": ["error", "true-never"]` }` suggestion that was made, but I'm a bit hesitant to do so given:

> But I'm not sure if it's common style enough to add into ESLint core.

I don't really know which metrics they use and I don't want to be pushy, so I'm not sure what the next steps are.

mightyiam:
Thank you for following up, @christianbundy.

ESLint does have an RFC process but this change seems to not qualify for that:
https://github.com/eslint/rfcs/blob/master/README.md

What I would do in this case is make a documentation-only PR for the `true-never` option and present it as a question to the maintainers, whether they find the approach appropriate.

Please be so kind as to update here on your progress.

feross:
FYI I'm not going to have time to look at this for several weeks.

PR #1407, opened 2019-09-09
chore: .npmrc with package-lock=false

mightyiam:


.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=false
```
mightyiam: https://docs.npmjs.com/misc/config#package-lock

christianbundy:
I'm a lockfile-lover, but since this repo doesn't use a lockfile (:scream:) I think it's a good idea to encode that preference in `.npmrc`.

:+1:

PR #1408, opened 2019-09-09
chore: mv eslintrc.json .eslintrc.json

mightyiam:


feross:
This seems like an edge case and I'd rather not rename this in case since it may break a downstream user who's requiring the file by its current name.

PR #1418, opened 2019-09-17
Don't run on Node.js versions less than 8.6.0

watson:
ESLint uses the spread operator which isn't supported in Node.js versions earlier than 8.6.0.

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

Previously the cli would just check that the major version was 8 or above before continuing. This change now also looks at the minor version to ensure that the full version is v8.6.0 or above.

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -1,7 +1,11 @@
 #!/usr/bin/env node
 
-if (process.version.match(/v(\d+)\./)[1] < 8) {
-  console.error('standard: Node 8 or greater is required. `standard` did not run.')
-} else {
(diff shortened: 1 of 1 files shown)


feross:
14.3.1

watson:
@feross Does the `14.3.1` comment mean that it will be released in v14.3.1, or that it has been released in that version? Because I don't see it on npm yet 🤔

feross:
> Does the 14.3.1 comment mean that it will be released in v14.3.1, or that it has been released in that version?

It means that I released it on npm. Except in this case, I didn't actually do the release! The build was failing so I went to fix that and forgot to actually publish. Fixed now!

PR #1432, opened 2019-10-03
Update README-ja.md for 3e6b299

munierujp:
Update README-ja.md to apply changes of original documents.

* 3e6b299


feross:
Thanks!

PR #1435, opened 2019-10-07
Update README-ja.md for d901c54

munierujp:
Update `README-ja.md` to apply changes of original documents.

* d901c54

As it is a simple English, no review by other Japanese speakers is required.
Please check only whether codes and commands are correct.


feross:
Thanks!

PR #1445, opened 2019-10-29
perf: make the rule 'No octal literals' more specific.

ZY2071:

[ X ] Documentation update
[ ] Bug fix
[ ] New feature
[ ] Other, please explain:

change the example for "No octal literals",  '042' => 34
NO
Maybe 34 is more specific here.


RULES.md
```diff
@@ -890,7 +890,7 @@ your code.
 
   ```js
   const num = 042     // ✗ avoid
-  const num = '042'   // ✓ ok
+  const num = 34   // ✓ ok
```
mightyiam: How about
```js
const octal = 042         // ✗ avoid
const decimal = 34        // ✓ ok
const octalString = '042' // ✓ ok
```

ZY2071: That's good. 👍

PR #1449, opened 2019-11-05
Added package.json funding property

ruyadorno:

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature
- [x] Other, please explain:

### What changes did you make? (Give an overview)
**npm** just added a new [`npm fund` command](https://blog.npmjs.org/post/188841555980/updates-to-community-docs-more) in it's [v6.13.0](https://github.com/npm/cli/releases/tag/v6.13.0) release as part of the efforts to help out the OSS community.

This PR adds `funding` info to **standard** so that users depending on it can actually retrieve the funding url when they run `npm fund` in their projects.

Note: The funding information will only be available once a new version of the package is published to the npm registry.

### Which issue (if any) does this pull request address?
Funding

### Is there anything you'd like reviewers to focus on?
N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,10 @@
   "engines": {
     "node": ">=8"
   },
+  "funding": {
+    "type": "Tidelift",
+    "url": "https://tidelift.com/subscription/pkg/npm-standard"
+  },
   "homepage": "https://standardjs.com",
   "keywords": [
     "JavaScript Standard Style",


package.json
```diff
@@ -36,6 +36,10 @@
   "engines": {
     "node": ">=8"
   },
+  "funding": {
+    "type": "Tidelift",
+    "url": "https://tidelift.com/subscription/pkg/npm-standard"
```
mightyiam: How did you come up with these values, please?

ruyadorno: I just clicked on the GitHub **Sponsor** button and followed the Tidelift url from there, feel free to ignore this PR and implement whatever makes sense for the project 😊 I just wanted to raise awareness about this new `package.json` property that is now supported by **npm**

mightyiam: Your PR is appreciated, @ruyadorno. As we do in programming, I would like to make sure that the values are correct. And I could not find the definitions for these fields. The best is the [rfc](https://github.com/npm/rfcs/blob/latest/accepted/0017-add-funding-support.md#implementation), in which there is no enumeration of the `type` field.

Also, following the `url` doesn't seem to let my easily donate. So I'm not sure it's the perfect URL. @feross would know.

ruyadorno: right right 👍 def sounds like it's not the best url to use in this case

feross:
I appreciate the thought behind this PR ❤️. I have a script that adds this information automatically to `package.json` upon publish now 👍

PR #1455, opened 2019-11-24
Add support for Node.js 8.3 - 8.6

LinusU:
The spread operator was actually introduced in Node.js 8.3. I think the typo comes from that node.green shows the latest version that has identical test results as earlier versions, and thus the column with green spread says "8.6.0", but hovering it shows it all the way down to "8.3.0".

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

Fixes a small error from #1418

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -4,8 +4,8 @@ var match = process.version.match(/v(\d+)\.(\d+)/)
 var major = parseInt(match[1], 10)
 var minor = parseInt(match[2], 10)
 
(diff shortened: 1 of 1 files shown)


mightyiam:
@LinusU is the CI failure related or not?

LinusU:
Hmm, nope, it shouldn't be related. Seems like it fails to git clone something? 🤔

mightyiam:
Hey, @LinusU: #1456. Up for grabs... :roll_eyes:

Issue #743, opened 2017-01-11
bundle.js ignored


dcousens:
I understand this is probably because most people use `bundle.js` for packaging... but oh god this took me 20 minutes of time wasting to figure out what was going wrong with my editor.

Do we need to special case this?
Doesn't everyone separate their source folders from distribution anyway?
In which case, if this is an issue, they can just do `standard src/`.

All the other defaults:

> node_modules/**, *.min.js, ~~bundle.js~~, coverage/**

I don't think would ever get in the way.

gismodo:
发自我的小米手机http://m.browser.baidu.com/



发自我的小米手机http://m.browser.baidu.com/



发自我的小米手机http://m.browser.baidu.com/
On Daniel Cousens <notifications@github.com>, Jan 11, 2017 19:14 wrote:

I understand this is probably because most people use bundle.js for packaging... but oh god this took me 20 minutes of time wasting to figure out what was going wrong with my editor.

Why do we need to special case this?
Doesn't everyone separate their source folders from distribution anyway?

feross:
Sorry you lost 20 minutes. That's always frustrating. 😢

I have lots of repos that generate `bundle.js` and I usually never have a `src/` folder for small projects. I personally appreciate having `bundle.js` as a default ignore. I've also never heard about anyone else running into this issue before.

That said, I agree that out of all the defaults `bundle.js` is the only one that could theoretically cause trouble if someone accidentally uses that name. The other ignores will almost never be a problem.

I think the usefulness outweighs the potential for confusion, given the prevalence of `bundle.js` but open to other people's thoughts.

dcousens:
@feross I know it would be a bit special case like,  but maybe don't ignore if the file is deliberately specified on the command line?

```
standard bundle.js
```

Just being a no-op was *especially* confusing,  but alas.

dcousens:
Or perhaps hint at what is ignored in a `verbose` mode?

feross:
@dcousens Yeah, these are both reasonable suggestions. Right now there's no way to override the built-in ignores.

gismodo:
发自我的小米手机http://m.browser.baidu.com/

dcousens:
@feross ha,  bitten by this again since I started using the beta (had previously removed it from my ignore globally installed version).

jsonmaur:
[...]

PR #1460, opened 2019-12-06
Update the list of default ignored files in documentation

ciffelia:

[X] Documentation update
[ ] Bug fix
[ ] New feature
[ ] Other, please explain:

Remove `bundle.js` from the list of of default ignored files, because it is no longer ignored. (https://github.com/standard/standard-engine/commit/1319c5b4d6b39eecf909a7072ae92eabe36e0213)

#743

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,7 +375,7 @@ automatically.
 
 ## How do I ignore files?
 
-Certain paths (`node_modules/`, `coverage/`, `vendor/`, `*.min.js`, `bundle.js`,
+Certain paths (`node_modules/`, `coverage/`, `vendor/`, `*.min.js`,
 and files/folders that begin with `.` like `.git/`) are automatically ignored.
 
 Paths in a project's root `.gitignore` file are also automatically ignored.
--- a/docs/README-esla.md
+++ b/docs/README-esla.md
@@ -367,7 +367,7 @@ automatically fix some problems.") cuando detecta errores que pueden ser arregla
 
 ## ¿Cómo hago para ignorar archivos?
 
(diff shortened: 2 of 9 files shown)


ciffelia:
@mightyiam This is the commit that removed `bundle.js`. https://github.com/standard/standard-engine/commit/1319c5b4d6b39eecf909a7072ae92eabe36e0213

mightyiam:
I don't understand how a single person can reasonably edit documentation in multiple languages. You don't know all of these languages — right, @ciffelia? Could the edit to one of these translations be erroneous?

How do we handle this? What's the strategy here?

ciffelia:
@mightyiam You're right. I thought removing a word from docs can be done without knowledge of languages, but actually I'm not sure. If you think this change should be reviewed by translation contributors, I will revert changes on translated docs.

mightyiam:
From my perspective, without a readily-available team of translators, this is unmaintainable. I suggest that authors modify only languages that they know and then if language contributors care, they will follow pull requests and contribute updates to documentation.

LinusU:
I personally think that it's probably fine, seems like a reasonable change in all languages 👍

mightyiam:
Yes, I took a close look and it seems fine.

mightyiam:
But languages have quirks that you wouldn't imagine.

ciffelia:
As @mightyiam pointed out, I can't guarantee the correctness of translations, so I'll revert changes on them for now.

ciffelia:
And as a Japanese speaker, I'm sure that the Japanese translation is correct.

PR #1461, opened 2019-12-08
Update Japanese documents to follow original documents

munierujp:
I've updated Japanese documents to to apply changes of original documents. (from e514626 to ecda198)
No review by other Japanese speakers is required.

## All original committ list
1. e514626
1. 4bc1671
1. ecda198 (no changes for Japanese document)


feross (approved):
LGTM

PR #1495, opened 2020-03-15
ci: test on the minimum required Node.js version

watson:
Because of #1494, this PR will fail for now. Once that issue has been fixed I'll rebase this PR to verify that all now works.

**What is the purpose of this pull request? (put an "X" next to item)**

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain: Improve CI

**What changes did you make? (Give an overview)**

Added Node.js v8.0.x to the Travis CI test matrix.

**Which issue (if any) does this pull request address?**

This ensures that issues like #1494 can't happen again.

**Is there anything you'd like reviewers to focus on?**

PR #1496, opened 2020-03-15
Don't run on Node.js versions less than 8.10.0

### LINKED ISSUE #1494: v14.3.2 broke support with Node.js versions <8.10.0
**What version of this package are you using?**

14.3.2

**What operating system, Node.js, and npm version?**

Node.js 8.9.4

**What happened?**

```
standard: Unexpected linter output:

./node_modules/eslint/lib/rules/no-octal-escape.js:41
                    /^(?:[^\\]|\\.)*?\\([0-3][0-7]{1,2}|[4-7][0-7]|0(?=[89])|[1-7])/su
                    ^

SyntaxError: Invalid regular expression flags
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:607:28)
    at Object.Module._extensions..js (module.js:654:10)
    at Module.load (module.js:556:32)
    at tryModuleLoad (module.js:499:12)
    at Function.Module._load (module.js:491:3)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
    at no-octal-escape (./node_modules/eslint/lib/rules/index.js:171:30)

If you think this is a bug in `standard`, open an issue: https://github.com/standard/standard/issues
```

It's complaining about the use of [the `s` regular expression operator](https://tc39.es/ecma262/#sec-get-regexp.prototype.dotAll) which wasn't introduced until Node.js 8.10.0.

**What did you expect to happen?**

That the linting completed successfully.

**Are you willing to submit a pull request to fix this bug?**

If it can help. That would involve rolling back eslint however to the previous version.

watson:
ESLint uses the `s` (dotAll) flag for regular expressions which isn't supported in Node.js versions earlier than 8.10.0.

I made a similar PR for when the spread operator was introduced in Node.js v8.6.0: #1418

Once this lands, I'll rebase #1495 to make sure we test early Node.js versions in CI.

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

I changed the minimum Node.js version where `standard` will run to be 8.10.0 (was 8.6.0).

closes #1494
closes #1455

Consider if this is actually a breaking change. If so we need to roll back the version of eslint instead of this. However, I think this be accepted so it's not breaking given the circumstances.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - 8.10.0
   - 8
   - 10
   - 12
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -4,8 +4,8 @@ var match = process.version.match(/v(\d+)\.(\d+)/)
 var major = parseInt(match[1], 10)
 var minor = parseInt(match[2], 10)
 
-if (major >= 9 || (major === 8 && minor >= 6)) {
+if (major >= 9 || (major === 8 && minor >= 10)) {
   require('standard-engine').cli(require('../options'))
 } else {
-  console.error('standard: Node 8 or greater is required. `standard` did not run.')
+  console.error('standard: Node 8.10.0 or greater is required. `standard` did not run.')
 }
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "tape": "^4.6.0"
   },
   "engines": {
-    "node": ">=8"
+    "node": ">=8.10.0"
   },
   "homepage": "https://standardjs.com",
   "keywords": [


watson:
My approach in #1495 doesn't work. Instead, I'll add the minimum version we think we should work on to the `.travis.yml` file in this PR. We could later consider adding a special test for 8.9.4 as well.

LinusU:
I think that this approach (minimum as indicated by `engines.node`, and latest of every stable major) is very good 👍 

That's how I do it in many of my packages as well :)

PR #1498, opened 2020-03-15
ci: test on latest non-supported Node.js version

watson:
Ensure that we don't fail on non-supported Node.js versions but just exit gracefully.

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain:

Added Node.js v8.9.x to the Travis CI test matrix with a special test script.

None

Verify that test new test for Node.js v8.9.x exits successfully while printing to STDERR that that particular version of Node.js isn't supported.


feross:
This is a great PR – sorry for taking so long to review. @watson Can you update it for the latest version of `standard` which supports 10.12.0 and greater. So perhaps the no-op test should be for 10.11.0?

feross:
I went ahead and updated the PR and merged it.

PR #1545, opened 2020-09-25
add Jublia to standard users list

logustra:
add Jublia to list of standard users
we are using JavaScript Standard Style for linting our front-end projects
thanks for the great tool!

I also updated the same section in other languages for you!


feross (approved):

feross:
thanks!

Issue #1590, opened 2020-11-04
Mention `ts-standard` in the docs 


feross:
I don't believe we're mentioning `ts-standard` in the docs. Isn't that the easiest way to use `standard` with TypeScript at the moment. Can someone – @toddbluhm @mightyiam – confirm this?

If so, can you please send a PR to update the TypeScript instructions in README.md.

LinusU:
`ts-standard` is definitely the way to go for TypeScript + Standard currently 👍

toddbluhm:
Yeah, agreed. I will update the docs, create new PR, and link it to this issue sometime this week.

toddbluhm:
Oh, before I go ahead and update the docs, do we want to move `ts-standard` to the `standard` org? @feross

feross:
@toddbluhm Sure! It seems like the best way to do `standard` + `typescript` :) Happy to bless it :)

toddbluhm:
@feross the transfer is complete https://github.com/standard/ts-standard. If I could get admin/owner access to it so I can finish setting up the coverage/dependency stuff that would be great 😄 Speaking of which do we have a standard coverage/dependency solution we want to use across projects?

feross:
Yep – we use Dependabot since it's built in to GitHub and does the job. It's currently only enabled on the `standard/standard` repo. I'd like to expand this to other repos.

I'm slightly negative on test coverage. For most `standard` repos which consist of sets of eslint rules, it's low value. If you want to enable it on `ts-standard`, go for it.

Just made you admin on `ts-standard`.

toddbluhm:
@feross Great! I was using Dependabot as well. I thought making me an admin would enable me to add Dependabot, but apparently, that is at an org level. If you could enable Dependabot for `ts-standard` that would be great. It should be all set up and ready good, just needs to be enabled.

For code coverage, I was using coveralls.io. No opinion either way, just what I happened to be using. I think you might need to enable `coveralls` as well. If you would like to use a different code coverage system that works for me as well. My project is already 100% covered and since it is that way I figure might as well show it. I really only do coverage for public projects to give devs increased confidence in the project (though admittedly it is easy to beat the system if you want to).

feross:
@toddbluhm Dependabot is enabled: 

<img width="957" alt="Screen Shot 2020-11-17 at 12 45 48 PM" src="https://user-images.githubusercontent.com/121766/99459862-9e181e00-28e3-11eb-9397-c3ca8f34126f.png">

[...]

PR #1595, opened 2020-11-10
docs: update typescript support documentation

toddbluhm:

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature
- [ ] Other, please explain:

Updated the typescript documentation to point to the officially supported typescript solutions.

#1590

Is this sufficient documentation? Since both projects are a bit complex I figured their README files
would be better reading than trying to include a lot of information in this help section. Also, is it fine
that I removed the old typescript solution since it seemed to be only a partial solution anyway?

### DIFF
--- a/README.md
+++ b/README.md
@@ -541,56 +541,16 @@ Or, add this to `package.json`:
 
 ### TypeScript
 
-To use TypeScript, you need to run `standard` with `@typescript-eslint/parser` as the parser,
-`@typescript-eslint/eslint-plugin` as a plugin, and tell standard to lint `**/*.ts` files (since it
-doesn't by default).
+There are two officially supported methods of using standard with typescript files.
 
-Unfortunately, there's an outstanding [issue](https://github.com/standard/standard/issues/1283)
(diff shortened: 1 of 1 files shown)


feross (approved):
LGTM

feross:
> Since both projects are a bit complex I figured their README files would be better reading than trying to include a lot of information in this help section
Can we reverse the order that TypeScript and Flow are referenced? At this point TypeScript is vastly more popular.

Also – since we're trying to provide an out-of-the-box experience, I'd de-emphasize the eslint config option a bit. A further improvement would be to at least include the "hello world" example for `ts-standard`, similar to how Flow includes a quick start example. At a minimum explaining `npm install ts-standard` and then running `ts-standard`. I'd mention at the end that if you want further control, you can use `eslint-config-standard-with-typescript`, similar to how we mention `eslint-config-standard` here: https://standardjs.com/#i-disagree-with-rule-x-can-you-change-it

toddbluhm:
@feross thanks for the feedback! I'll update the docs based on what you have said.

PR #1597, opened 2020-11-10
Support node 11

alanclarke:

[x] Bug fix

Currently, if you try to run standard using node v11, the executable outputs: `Node 10.12.0 or greater is required. standard did not run.`

Since v11 is greater than v10.12.0, this change makes the behaviour consistent with the messaging!

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -5,7 +5,7 @@ var match = process.version.match(/v(\d+)\.(\d+)/)
 var major = parseInt(match[1], 10)
 var minor = parseInt(match[2], 10)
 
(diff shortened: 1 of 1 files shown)


LinusU:
We don't want to support Node.js 11 since it's no longer supported by Node.js itself, can you instead update the message?

feross:
@LinusU While I agree that we don't want to *officially* promise support for Node 11, is there harm in letting it run in on that version if it works?

LinusU:
Right, didn't really think about that 😄 

I don't see why we should block it 👍

LinusU (approved):

feross:
Released as 16.0.2

PR #1629, opened 2021-01-26
Use GitHub Actions instead of Travis CI

HonkingGoose:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain: Migrate from Travis CI to GitHub Actions

- Removed travis.yml configuration file
- Created 2 GitHub Action config files (one for recent node versions, one for ancient Node version)
- Add `github-actions` to list of things that Dependabot will update

In general this should work the same as what you're used to on Travis CI.
The tests will run on each push to `master`, and on each pull request that targets `master`.

You might need to tinker with your GitHub branch protection settings after merging this PR.

After you've merged this PR, I'll make a PR to add new testing status badges to the README.md file (the old badge still points to Travis for now).

Travis CI is migrating from a `.org` to a `.com` domain.
If you're running tests on the `.org` domain, those will stop working soon.

If you migrate to their free plan on the `.com` domain you'll get a one-time 1000 free build minutes.
Once the minutes are used up you don't get any more without paying or applying to their OSS plan.

Quote from https://www.travis-ci.com/plans:

> Free for open source. We love the Open Source Community, and to show how much we love it, upon validated request placed with our Support Team you may receive free OSS credits for your public builds.

Quote from https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing:

> 
> - We will be offering an allotment of OSS minutes that will be reviewed and allocated on a case by case basis. Should you want to apply for these credits please open a request with Travis CI support stating that you’d like to be considered for the OSS allotment. Please include:
>   - Your account name and VCS provider (like `travis-ci.com/github/[your account name]` )
>   - How many credits (build minutes) you’d like to request (should your run out of credits again you can repeat the process to request more or discuss a renewable amount)

This sounds like a lot of hassle, and even if you jump through the hoops and get credits now, there's no guarantee that you'll keep getting the needed amount of credits in the future.

Given all this, I think it's better to just migrate to GitHub Actions, as that is free for public repositories, full stop.

I have pinned the ancient node version to `0.10.48`, which is the latest release of the `0.10.x` line.
I could not get the GitHub Action to work with `0.10.0`.

My work is based off the GitHub actions starter workflow for Node JS: https://github.com/actions/starter-workflows/blob/main/ci/node.js.yml, this is licensed under the MIT license.


HonkingGoose:
I notice that running GitHub actions is blocked on this repository, you can see the tests in action on my fork: https://github.com/HonkingGoose/standard/pull/2

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,24 @@
...
+      matrix:
+        node-version: [10.12.0, 10.x, 12.x, 14.x]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: npm install
```
Divlo, line 23: We could cache dependencies to have better performances in CI, something like that :
```yaml
- name: 'Cache dependencies'
  uses: 'actions/cache@v2'
  with:
    path: |
      **/node_modules
    key: ${{ runner.os }}-${{ hashFiles('**/package.json') }}
```

.github/workflows/old-node.js.yml
```diff
@@ -0,0 +1,30 @@
...
+      matrix:
+        node-version: [0.10.48]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: npm install
```
Divlo, line 27: Same thing, we could cache dependencies.

Also in my opinion, it is not necessary to test if we don't support a version, since we don't support, if you know what I mean, the user should see what version it supports, and install the correct one, so maybe we don't need to have this extra CI check, so we could have faster CI, maybe I'm wrong, just my thought.

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,24 @@
...
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [10.12.0, 10.x, 12.x, 14.x]
```
Divlo, line 15: I don't know why in the `.travis.yml` there was `10.12.0` but is that so different with `10.x` ?
Maybe we don't need it, and we could only run in `10.x`.
so `node-version: [10.x, 12.x, 14.x]`.

Divlo (changes requested):
Thanks for your contribution! @HonkingGoose

I definitely agree that we should change `travis` to `GitHub Actions`. :+1:
I also opened a PR in `standard-engine` to do that : https://github.com/standard/standard-engine/pull/233

> After you've merged this PR, I'll make a PR to add new testing status badges to the README.md file (the old badge still points to Travis for now).

I guess you could already change the badges in the README even if it is broken currently because there are not GitHub Actions yet, it doesn't matter, once this PR has been merged, the badges should work as expected.

Also see my review comments. :eyes:

.github/workflows/node.js.yml
HonkingGoose: The `actions/cache` project does not recommend that we use `**/node_modules`.
They recommend we do this:

```yaml
- uses: actions/cache@v2
  with:
    path: ~/.npm
    key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
    restore-keys: |
      ${{ runner.os }}-node-
```

Source: https://github.com/actions/cache/blob/main/examples.md#node---npm

.github/workflows/node.js.yml
HonkingGoose: I also don't find pinning to 10.12.0 very useful, so I've removed the pinned version.
The tests still run on 10.x anyways.

.github/workflows/node.js.yml
HonkingGoose: And the GitHub docs recommend this:

> This example creates a new cache when the packages in `package-lock.json` file change, or when the runner's operating system changes.

```yaml
    - name: Cache node modules
      uses: actions/cache@v2
      env:
        cache-name: cache-node-modules
      with:
        # npm cache files are stored in `~/.npm` on Linux/macOS
        path: ~/.npm
        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
        restore-keys: |
          ${{ runner.os }}-build-${{ env.cache-name }}-
          ${{ runner.os }}-build-
          ${{ runner.os }}-
```

Source: https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action

.github/workflows/old-node.js.yml
HonkingGoose: The `0.10.48` CLI output test is really fast it's done in about a minute.
The full-fat Node 10, 12 and 14 tests take anywhere between 7 and 10 minutes.

So we don't need to drop the old Node test for fear of having to wait around on that test. 😄

HonkingGoose:
## About caching without a lockfile

I think using caching without having a `package-lock.json` file is going to cause problems sooner or later...
As it stand now, I can only tell GitHub action to refresh the cache whenever the `package.json` file is changed.

If a peer-dependency or transient dependency updates, it will not change the `package.json` file. (A `package-lock` file would definitely change!). Now the GitHub Action runner is using old stuff!

## If you want to cache, use a lock-file

I think caching is only a good idea if you're actually using a lock-file.
This will also have the added benefit of making the build reproducible.

If you're scared of having to deal with package-lock updates manually, you could consider migrating to Renovate bot instead, and update the lockfile via a automated PR once a week.

Or you can just not use the cache functionality. 😄 
That **might** slow down the tests a bit, but you're not messing around with a lock file either.

I'll leave you to decide how to weigh this.

Divlo (approved):

Divlo:
Great! :+1: 

I guess we'll cache dependencies, personally I always cache dependencies for the CI on my own repo, and I never ran into issues or bugs, maybe I'm lucky enough, and I've got faster CI.

> I think using caching without having a package-lock.json file is going to cause problems sooner or later...

I don't see anywhere in the docs, that we should not cache dependencies without `package-lock.json`.

That's another discussion not directly related to this PR but in my opinion, if we everyone could use exact version in their `package.json` and we didn't invented version match with `^` or `~` etc symbol, we would have less problems.
In my own repo even if it's not an npm package, I only use `package.json` and ignore `lockfiles` so it works with whatever package manager `npm`, `yarn`, `pnpm`, etc, and we don't need to manage a big ugly `package-lock.json` file, also in `git` we should avoid to commit generated files, as we do for build.

> This will also have the added benefit of making the build reproducible.

That's right, but if every packages would use exact version match, we would not need to use `package-lock.json`, I know that's not the case unfortunately.

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,38 @@
+name: Node CI
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
```
munierujp, line 10: I think `test` is a better name for these jobs.

munierujp (commented):
See my comments.

.github/workflows/old-test.yml
```diff
@@ -0,0 +1,43 @@
...
+name: Old test
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
```
HonkingGoose, line 14:
```suggestion
  test:
```

Or did you mean this?

.github/workflows/test.yml
```diff
@@ -0,0 +1,38 @@
+name: Test
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
```
HonkingGoose, line 10:
```suggestion
  test:
```

Or did you mean this?

.github/workflows/old-test.yml
munierujp: Both.
As you know, the job names will be displayed on pages.

スクリーンショット 2021-01-29 23 32 30

.github/workflows/test.yml
munierujp: see https://github.com/standard/standard/pull/1629#discussion_r566862061

.github/workflows/old-test.yml
HonkingGoose: I've changed this now. 😄

HonkingGoose:
Also, do you want Dependabot to check for GitHub Action runner updates?

Right now you won't get any notifications if the action runners get updates from `v2` to `v3`.

Dependabot opens a PR for each minor version though (and even alpha releases..), so this can get spammy...

The dependabot.yml file would get a addition like this:

```yml
# Set update schedule for GitHub Actions

version: 2
updates:

  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      # Check for updates to GitHub Actions every weekday
      interval: "daily"
```

GitHub docs: https://docs.github.com/en/github/administering-a-repository/keeping-your-actions-up-to-date-with-dependabot#about-dependabot-version-updates-for-actions

Divlo:
Yes, sure, you could add `dependabot` for GitHub Actions. :+1: @HonkingGoose

.github/dependabot.yml
```diff
@@ -8,3 +8,10 @@ updates:
       - "dependency"
     # Always increase the version requirement to match the new version.
     versioning-strategy: increase
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - "github-actions"
```
HonkingGoose, line 17: Dependabot will label GitHub Actions PRs with the `github-actions` label.
That way you can see by the label that it's different from a normal `dependency` update.

Divlo (approved):

HonkingGoose:
It might be a good idea to limit what GitHub Actions are allowed to run on this repository.
It's possible for users to run arbitrary actions with a malicious PR.

I recommend that the repository admin reads [GitHub docs, disabling or limiting GitHub Actions for a repository](https://docs.github.com/en/github/administering-a-repository/disabling-or-limiting-github-actions-for-a-repository).

munierujp (approved):
LGTM 😄

HonkingGoose:
Thanks for merging in my work! I hope you enjoy your new CI. 😉

PR #1634, opened 2021-02-02
ci: add automated release on npm : `npm-publish`

Divlo:

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature
- [x] Other, please explain: Add automated release process on `npm`

- Create a GitHub Action to automatically release the package on npm when someone create a new release on GitHub (see `.github/workflows/npm-publish.yml)`
You will need to set a new secret `NPM_TOKEN` on the repo so it could works as it is [explained there](https://docs.github.com/en/free-pro-team@latest/actions/guides/publishing-nodejs-packages).

Also once this has been merged, I'll open smiliar PRs to `ts-standard`, `standard-engine`,  `eslint-config-standard` and `eslint-config-standard-with-typescript`.

Why is this important ? It's much better to have automated deployment, also sometime, you publish to npm but you forgot to create the release on GitHub, so now the releases would be synchronized and it's automated.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,37 @@
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: 'Publish package to npm'
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  publish-npm:
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: 'actions/checkout@v2'
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+          registry-url: 'https://registry.npmjs.org/'
+
+      - name: 'Cache Node dependencies'
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+
+      - name: 'Install dependencies'
+        run: 'npm install'
+
+      - name: 'Run tests'
+        run: 'npm test'
+
+      - name: 'Publish to npm registry'
+        run: 'npm publish --access public'
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}


Divlo:
Maybe you could add your review on it. @HonkingGoose :eyes: 

Also once this has been merged, I'll open smiliar PRs to `ts-standard`, `standard-engine`,  `eslint-config-standard` and `eslint-config-standard-with-typescript`.

Why is this important ? It's much better to have automated deployment, also sometime, you publish to npm but you forgot to create the release on GitHub, so now the releases would be synchronized and it's automated.

We could also add `commitlint` and add automated generated changelog thanks to commits names, I didn't implemented that yet on this PR, but feel free to discuss about it there, and share with us your thoughts.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,31 @@
...
+on:
+  release:
+    types: [created]
+
+jobs:
+  publish-npm:
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: 'actions/checkout@v2'
+
```
HonkingGoose, line 14:
```suggestion
  publish-npm:
    runs-on: 'ubuntu-latest'
    
    strategy:
      matrix:
        node-version: [14.x]

    steps:
      - uses: 'actions/checkout@v2'

```

I think you need to define `matrix.node-version` if you want to use it in `${{ matrix.node-version }}`. 😉

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,31 @@
...
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+
+      - run: 'npm install'
+      - run: 'npm publish --access public'
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```
HonkingGoose, line 31:
```suggestion
      - name: Install dependencies
        run: 'npm install'
      
      - name: Run tests
         run: npm test
        
      - name: Publish to npm registry
         run: 'npm publish --access public'
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```

I think it's a good idea to stop the publish action if the tests fails. That way you're less likely to publish a broken release.
I also added `name:` fields to describe each step.
I don't know why the formatting is off.

You should probably do a `npx prettier --write .github/workflows/npm-publish.yml` on this branch after committing my suggestions.

HonkingGoose (changes requested):

.github/workflows/npm-publish.yml
Divlo: By default `actions/setup-node@v2` will use the latest LTS version of Node.js, and we only want to release once, so it makes sense to only run the action on a single version of `Node.js`, and to don't use `matrix.node-version`.

HonkingGoose: Oh okay, so you don't _need_ to specify a version when using `actions/setup-node@v2` then... 😄 Well I learned something new. 😄

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,37 @@
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: 'Node.js Package'
```
HonkingGoose, line 3:
```suggestion
name: 'Publish npm package'
```

Maybe something like this is clearer?

Divlo: Indeed, actually I just took the example from GitHub Actions docs. :smile:

.github/workflows/npm-publish.yml
Divlo: Oh my bad, actually I checked again but you were right, excuse me for misleading you.

In fact, I copied the config file from https://github.com/Divlo/create-fullstack-app/ that was also copied from the docs, but I certainly not on purpose deleted these 2 lines :
```yaml
node-version: 14
registry-url: 'https://registry.npmjs.org/'
```

As it is working for `create-fullstack-app`, I thought, it would also working there, but I forgot to copy these 2 lines so if it has not the same configuration he could not work the same ah ah. :laughing: 

But it is better to use only 1 version to release, so I don't think we need to still, declare `strategy.matrix` etc.
What do you think ? @HonkingGoose

HonkingGoose: > Oh my bad, actually I checked again but you were right, excuse me for misleading you.

Oh well, we all make mistakes.. 😄 
So we **do need** to define at least one Node version.

> But it is better to use only 1 version to release, so I don't think we need to still, declare strategy.matrix etc.
What do you think ? @HonkingGoose

I agree, just use 1 version to create the release.
I think if you ran this action with a Node version matrix, you'd probably make multiple releases (1 release for Node v12, 1 release for Node v14 etc.)

HonkingGoose:
@Divlo Have you checked out the `semantic-release` bot, it might be what you're looking for...
You can find it here: https://github.com/semantic-release/semantic-release

Quote from their README:

> ### Release steps
> 
> After running the tests, the command `semantic-release` will execute the following steps:
> 
> | Step              | Description                                                                                                                     |
> | ----------------- | ------------------------------------------------------------------------------------------------------------------------------- |
> | Verify Conditions | Verify all the conditions to proceed with the release.                                                                          |
> | Get last release  | Obtain the commit corresponding to the last release by analyzing [Git tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging). |
> | Analyze commits   | Determine the type of release based on the commits added since the last release.                                                |
> | Verify release    | Verify the release conformity.                                                                                                  |
> | Generate notes    | Generate release notes for the commits added since the last release.                                                            |
> | Create Git tag    | Create a Git tag corresponding to the new release version.                                                                      |
> | Prepare           | Prepare the release.                                                                                                            |
> | Publish           | Publish the release.                                                                                                            |
> | Notify            | Notify of new releases or errors.                                                                                               |

I've never used the  `semantic-release` bot, so I cannot help with configuring it.
I think you can ask for help in the discussion tab on the linked repository.

They have a "recipe" for getting the bot setup with GitHub Actions: https://github.com/semantic-release/semantic-release/blob/master/docs/recipes/github-actions.md

Divlo:
Yes, I already know the `semantic-release` bot, it's great but it seems like it is only working with git branches to trigger a release...
I mean as far as I know, I never used it.
It would be better to trigger a release with a git tag/release as it is explained in the description of this PR.

I would rather would like to use : [release-it](https://www.npmjs.com/package/release-it)
It is what I use for my personnal projects and it is working fine. 
You can see how I use it in [create-fullstack-app](https://github.com/Divlo/create-fullstack-app/).

But we'll need more opinions to decide, any thought @standard/team ?

ungoldman:
Not sure if standard uses this in practice much but if one were to create a beta release (e.g. `vX.Y.Z-rc1`, `vX.Y.Z-beta`), one would _not_ want that to be published with the default `latest` tag to npm. It doesn't look like this github action config accounts for that possibility.

Divlo:
You're right, no `standard` repo are using automated deployment yet, but that doesn't mean, it doesn't need to change.

As I said in the description of the PR once this has been merged, we'll try to also add automated deployment to others `standard` repo where possible, it will improve by far the velocity of releases.

Actually, with the changes of this PR, we can release `standard@16.0.3-beta` for example, to be clear this GitHub Action only install dependencies, run tests and run `npm publish` and npm publish look the version inside `package.json`.

This GitHub Action really does, what we do ourselves on our local machine to release a new version, nothing more.

It doesn't care about the release title on GitHub but in my opinion we should name the release on GitHub, the same as release version on `npm` so for example `v16.0.3-beta`.  @ungoldman

ungoldman:
I think you may not know about the case I'm talking about: `npm publish` publishes by default to the `latest` dist tag. Often maintainers may need to publish beta releases with a different dist tag so any regressions or breaking changes don't make it to a large user base. This is done by running, for example, `npm publish --tag=beta` for a release with a `beta` dist tag.

You can see some different dist tags in practice by running `npm info react` or `npm info express` -- at the bottom of the output you can see various dist tags.

```
dist-tags:
latest: 4.17.1       next: 5.0.0-alpha.8
```

The github action you have configured here would, as far as I understand, publish _any_ release detected in the repo under the `latest` dist tag, which would be a big problem if any maintainer is trying to use an alternate distribution tag for early testing.

Divlo:
Yes, I know the use case and what you were talking about, thanks anyway for the explanation. :smile: 

I thought you only needed to change the version number in `package.json` so it can release on another tag, but you're right as it is in the doc, we also need to run `npm publish --tag=beta` for example, it doesn't automatically, release on tag specified in the version number inside `package.json`.

In my opinion, it should be the case, and maybe it is, I ran `npm publish --dry-run` and here is what I've got :


I don't know what means `standard@16.0.3-beta1`, is it still released under latest tag with this version number ?
Anyway, I'll try to find a solution, and I'll keep you updated about what I've got.

feross:
I haven't ever used automatic publishing tools like this before and I'm hesitant to do it. I don't want to accidentally trigger a release with broken code or publish a major version too early. There's so many different packages that need to be updated when I do a new major release of `standard` and I'm already spending a lot of time on my local machine doing that work, so I don't see how saving a few seconds to run publish is very helpful for me here.

For patch and minor releases in-between major versions, I see how this could be useful, but we don't have many of those. Usually it's just every 3-6 months, I'll sit down for a day, merge in 20 new rules, and make a new major version release. Users actually value the stability and lack of frequency of `standard` releases.

I'm not saying I'm completely against this idea. It's just that I haven't tried it before and my default position is extremely skeptical that it will justify the extra complexity and uncertainty around when/if releases have gone out, accidental releases, wrong version releases, etc.

Divlo:
I understand your point of view, in fact I completely understand :
> spending a lot of time on my local machine doing that work

Thanks for doing this work! :tada: @feross 

That's why automated deployment could be useful and there is probably **something wrong with this old way** to do, because as developers we should spend more time making new features, fixing bugs instead of publishing packages and doing extra maintenance things.

> so I don't see how saving a few seconds to run publish is very helpful for me here.

It's not only few seconds, we could also imagine doing automated changelog, and all the extra things you do like running `tools/update-authors.sh` to update authors etc. To be honest I don't really know what you're doing when you're publishing a new release for `standard`, but the main idea is : it takes too much time and effort to maintain, right ?

>  I don't want to accidentally trigger a release with broken code or publish a major version too early
...
extra complexity and uncertainty around when/if releases have gone out, accidental releases, wrong version releases, etc

After all, we're humans, we make mistakes, but a machine only do the mistake that a humain ordered to do, but if the human from the start correctly programmed the machine, the machine would less likely makes mistake than an Human.
For example: A calculator is never wrong for doing calculations, but a human could be mistaken. Hopefully you understand my point, so to sum up, automated deployment will cause less errors than a human could do.

For example, I noticed that for the latest releases, you have forgotten to create the releases on GitHub, currently we're at `16.0.3`, but on GitHub, it is only `14.3.4`, that's a perfect example of why this could be useful, since as it is in this PR, you need to create a release on GitHub to trigger a release.

Many famous `npm` packages are also using automated deployment, and if you don't want that it is too much automated, currently the only thing that does this PR : when we create a release on GitHub, it runs `npm publish`, so there is not so much that could go wrong, since you're running the same command in your local machine, it does exactly the same thing.

> Users actually value the stability and lack of frequency of `standard` releases.

I agree, since `standard` is a linter/formatter, it should not change rules too much between releases, and it is stable enough, so we don't need too much update.

> I'm not saying I'm completely against this idea. It's just that I haven't tried it before

Novelty is always scary, I know, I was also scary to learn TypeScript, when I only knew plain JavaScript, but now that I learnt to use TypeScript, I do everything with it, I don't even write 1 line of JavaScript since then, and I don't even think that I should comeback to plain JavaScript, it is just so much better with TypeScript. :smile:

HonkingGoose:
You might also want to check out "release-please as a way to automate the release process, but keep manual control of when to release. To use release-please you need to follow the Conventional Commits style when making commits.

Quote from the "release-please" readme:

> ## What's a Release PR?
> 
> Rather than continuously releasing what's landed to your default branch,
> release-please maintains Release PRs:
> 
> These Release PRs are kept up-to-date as additional work is merged. When you're
> ready to tag a release, simply merge the release PR. Both squash-merge and
> merge commits work with Release PRs.
> 
> When the Release PR is merged, release-please takes the following steps:
> 
> 1. Updates your changelog file (for example `CHANGELOG.md`), along with other language specific files (for example `package.json`).
> 2. Tags the commit with the version number
> 3. Creates a GitHub Release based on the tag
> 
> You can tell where the Release PR is its lifecycle by the status label on the
> PR itself:
> 
> - `autorelease:pending` is the initial state of the Release PR before it is merged
> - `autorelease:tagged` means that the Release PR has been merged and the release has been tagged in GitHub
> - `autorelease:published` means that a GitHub release has been published based on the Release PR (_release-please does not automatically add this tag, but we recommend it as a convention for publication tooling_).

Learn more at the release-please GitHub repository: https://github.com/googleapis/release-please

voxpelli:
I'm also intrigued by `release-please`.

I also think it would be good to get the `authors` file update included in such a flow or in something close to that goal. Right now it's manual.

Divlo:
This PR has no recent activity, and probably won't be merged for a long time and as I don't want/have time to do the job for this PR, I decided to close it.

Feel free to open another PR yourself based on this one if you want this to be implemented.

Issue #1663, opened 2021-05-21
Readme badge announces that test is failing


matkoniecz:
**What version of this package are you using?**

Latest master

**What operating system, Node.js, and npm version?**

Irrelevant

**What happened?**

https://github.com/standard/standard#------javascript-standard-style----

![screen02](https://user-images.githubusercontent.com/899988/119100252-0c1ada80-ba18-11eb-8c27-f60f9136aed8.png)

**What did you expect to happen?**

No failing tests,

**Are you willing to submit a pull request to fix this bug?**

No.

Divlo:
Indeed, the tests are currently failing.
We need to investigate why and where it is coming from. :+1:

voxpelli:
I think it would be good to separate the GitHub Action that validates the validity of this repo on its own from the GitHub Action that validates against external repositories.

That way we could have two badges and more clearly indicate when the source of an error could possibly be external rather than internal.

Thoughts @Divlo?

It would require splitting the `tape` call here into two: https://github.com/standard/standard/blob/093b05608589e9345535db1f7d61cb5c4c006829/package.json#L71

Creating two new npm scripts, eg. `test-external` + `test-internal`, and rewiring `npm test` to be `npm run test-internal && npm run text-external` and have the current GitHub Action run the `test-internal` only and add a new GitHub Action that runs the `test-external` + add a new badge for the external repos.

That way the badging situation will be easier to decipher – for maintainers as well as users and contributors.

Divlo:
Sounds good to me! @voxpelli 
I will open a PR for that, I also think that we should remove the "Old test" GitHub Action as it is useless in my opinion, I will open another PR to remove this old test action, so we can merge/discuss both changes separately.

PR #1675, opened 2021-06-13
chore: remove typescript files from .pre-commit-hooks.yaml

sid-kap:
Since typescript syntax is not supported by the linter, by default IMO we should not lint these files.

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

Remove `.ts`/`.tsx` files from the pre-commit lint hook

In my mixed JavaScript/TypeScript project that uses pre-commit, I'm getting a syntax error (`Parsing error: Unexpected token :`) when `standard` tries to lint a TypeScript file

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,6 +4,5 @@
   language: node
   files: "\\.(\
     js|jsx\
-    |ts|tsx\
     |mjs|cjs\
     )$"


Divlo (approved):

sid-kap:
@Divlo looks like the same tests are also broken on master. Would it be okay to just merge this without fixing the tests in that case?

Divlo:
> @Divlo looks like the same tests are also broken on master. Would it be okay to just merge this without fixing the tests in that case?

There is an open issue about the tests failing (see: #1663).
I guess it would be okay to merge this even if the tests are actually failing, as it is not related yes. 

Keep in mind that, this will not solve your problem : `I'm getting a syntax error (Parsing error: Unexpected token :) when standard tries to lint a TypeScript file`, if you are actually using TypeScript, you should take a look at [ts-standard](https://github.com/standard/ts-standard), if you are only linting JavaScript files, you should edit your own `.pre-commit-hooks.yaml` file, this PR will not affect in any way your project.

@sid-kap

sid-kap:
Thanks!
Ah I see now, I didn't realize that ts-standard lints both typescript and javascript files. Anyway, I guess this will be useful in case someone wants to use `standard` for js and a different linter for ts files.

PR #1681, opened 2021-06-29
test: remove old Node.js version testing

Divlo:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain: Tests

Remove the test for the legacy Node.js version, I find it useless to test that we don't run in older versions of Node.js.
The user should be responsible to run old versions.

Instead what we can do is to clearly state in `README.md` what Node.js versions we support, that should be `>=12.0.0` as now v10 is EOL, maybe we should wait for the next major version of `standard` to remove Node v10 from the support list.

Thoughts?

Ref: https://github.com/standard/standard/issues/1663#issuecomment-870581010


voxpelli:
I'm actually 👎 on this change after reading the description of the test:

```
# Special test for the oldest version of Node.js that we "support"
# even though the linter won't actually run. Test that the command
# line program exits cleanly.
```

The intent of the test is to ensure that we fail silently on non-supported old node.js versions, thus making it simple for projects that supports a wider version range than us to still include us in their test script on all node versions.

(It could however do with some more added versions, as the jump from `0.10` to `<10.12` is quite large, just testing on `0.10` doesn't ensure that it fails silently on eg `6.x`)

voxpelli (changes requested):
Forgot to submit my review as a review

Divlo:
Closing.
It's probably better to have more than necessary automated tests than not enough.

PR #1707, opened 2021-09-13
feat: usage of ESM imports instead of CommonJS

jimmywarting:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain:

- Switch from cjs to esm.
- bumped lowest supported node version

standard-engine is holding back on some esm stuff...

> it would offer a smoother transition to migrate to ESM only for packages like standard, ts-standard etc, before migrating this one (standard-engine). https://github.com/standard/standard-engine/issues/251#issuecomment-913229242

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -1,13 +1,13 @@
 #!/usr/bin/env node
-/* eslint-disable no-var */
+/* eslint-disable no-var, no-eval */
 
 var match = process.version.match(/v(\d+)\.(\d+)/)
 var major = parseInt(match[1], 10)
 var minor = parseInt(match[2], 10)
 
 if (major >= 12 || (major === 12 && minor >= 20)) {
-  import('standard-engine').then(function (standardEngine) {
-    import('../options.js').then(function (options) {
+  eval('import("standard-engine")').then(function (standardEngine) {
+    eval('import("../options.js")').then(function (options) {
       standardEngine.cli(options.default)
     })
   })
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -5,8 +5,12 @@ var match = process.version.match(/v(\d+)\.(\d+)/)
 var major = parseInt(match[1], 10)
 var minor = parseInt(match[2], 10)
 
-if (major >= 11 || (major === 10 && minor >= 12)) {
-  require('standard-engine').cli(require('../options'))
+if (major >= 12 || (major === 12 && minor >= 20)) {
+  import('standard-engine').then(mod => {
+    import('../options.js').then(({ default: options }) => {
+      mod.cli(options)
+    })
+  })
 } else {
-  console.error('standard: Node 10.12.0 or greater is required. `standard` did not run.')
+  console.error('standard: Node 12.20.0 or greater is required. `standard` did not run.')
 }
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -6,9 +6,9 @@ var major = parseInt(match[1], 10)
 var minor = parseInt(match[2], 10)
 
 if (major >= 12 || (major === 12 && minor >= 20)) {
-  import('standard-engine').then(mod => {
-    import('../options.js').then(({ default: options }) => {
-      mod.cli(options)
+  import('standard-engine').then(function (standardEngine) {
+    import('../options.js').then(function (options) {
+      standardEngine.cli(options.default)
     })
   })
 } else {
--- a/index.js
+++ b/index.js
@@ -1,7 +1,5 @@
 /*! standard. MIT License. Feross Aboukhadijeh <https://feross.org/opensource> */
 import engine from 'standard-engine'
-import opts from './options.js'
+import options from './options.js'
 
-const Linter = engine.linter
-
-export default new Linter(opts)
+export default new engine.linter(options)
--- a/index.js
+++ b/index.js
@@ -1,5 +1,7 @@
 /*! standard. MIT License. Feross Aboukhadijeh <https://feross.org/opensource> */
-const Linter = require('standard-engine').linter
-const opts = require('./options')
+import engine from 'standard-engine'
+import opts from './options.js'
 
-module.exports = new Linter(opts)
+const Linter = engine.linter
+
+export default new Linter(opts)
--- a/index.js
+++ b/index.js
@@ -2,4 +2,6 @@
 import engine from 'standard-engine'
 import options from './options.js'
 
-export default new engine.linter(options)
+const Linter = engine.linter
+
+export default new Linter(options)
--- a/options.js
+++ b/options.js
@@ -1,8 +1,9 @@
 import { readFileSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
 import eslint from 'eslint'
 
 // eslintConfig.configFile have problem reading URLs and file:///
-const configFile = new URL('./eslintrc.json', import.meta.url).toString().slice(7)
+const configFile = fileURLToPath(new URL('./eslintrc.json', import.meta.url))
 const pkgURL = new URL('./package.json', import.meta.url)
 const pkgJSON = readFileSync(pkgURL, { encoding: 'utf-8' })
 const pkg = JSON.parse(pkgJSON)
--- a/options.js
+++ b/options.js
@@ -1,15 +1,21 @@
-const eslint = require('eslint')
-const path = require('path')
-const pkg = require('./package.json')
+import { dirname, join } from 'node:path'
+import { readFileSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
+import eslint from 'eslint'
 
-module.exports = {
-  bugs: pkg.bugs.url,
+const __dirname = dirname(fileURLToPath(import.meta.url))
+const json = readFileSync('./package.json', 'utf8')
+const pkg = JSON.parse(json)
+const { bugs, homepage, version } = pkg
+
+export default {
+  bugs: bugs.url,
   cmd: 'standard',
   eslint,
   eslintConfig: {
-    configFile: path.join(__dirname, 'eslintrc.json')
+    configFile: join(__dirname, 'eslintrc.json')
   },
-  homepage: pkg.homepage,
+  homepage: homepage,
   tagline: 'Use JavaScript Standard Style',
-  version: pkg.version
+  version: version
 }
--- a/options.js
+++ b/options.js
@@ -1,7 +1,8 @@
-import { join } from 'node:path'
 import { readFileSync } from 'node:fs'
 import eslint from 'eslint'
 
+// eslintConfig.configFile have problem reading URLs and file:///
+const configFile = new URL('./eslintrc.json', import.meta.url).toString().slice(7)
 const pkgURL = new URL('./package.json', import.meta.url)
 const pkgJSON = readFileSync(pkgURL, { encoding: 'utf-8' })
 const pkg = JSON.parse(pkgJSON)
@@ -11,7 +12,7 @@ export default {
   cmd: 'standard',
   eslint,
   eslintConfig: {
-    configFile: join(__dirname, 'eslintrc.json')
+    configFile
   },
   homepage: pkg.homepage,
   tagline: 'Use JavaScript Standard Style',
--- a/options.js
+++ b/options.js
@@ -1,21 +1,19 @@
-import { dirname, join } from 'node:path'
+import { join } from 'node:path'
 import { readFileSync } from 'node:fs'
-import { fileURLToPath } from 'node:url'
 import eslint from 'eslint'
 
-const __dirname = dirname(fileURLToPath(import.meta.url))
-const json = readFileSync('./package.json', 'utf8')
-const pkg = JSON.parse(json)
-const { bugs, homepage, version } = pkg
+const pkgURL = new URL('./package.json', import.meta.url)
+const pkgJSON = readFileSync(pkgURL, { encoding: 'utf-8' })
+const pkg = JSON.parse(pkgJSON)
 
 export default {
-  bugs: bugs.url,
+  bugs: pkg.bugs.url,
   cmd: 'standard',
   eslint,
   eslintConfig: {
     configFile: join(__dirname, 'eslintrc.json')
   },
-  homepage: homepage,
+  homepage: pkg.homepage,
   tagline: 'Use JavaScript Standard Style',
-  version: version
+  version: pkg.version
 }
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "tape": "^5.3.1"
   },
   "engines": {
-    "node": ">=12.20.0"
+    "node": "^12.20.0 || ^14.13.1 || >=16.0.0"
   },
   "homepage": "https://standardjs.com",
   "keywords": [
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "bin": {
     "standard": "bin/cmd.js"
   },
+  "type": "module",
   "bugs": {
     "url": "https://github.com/standard/standard/issues"
   },
@@ -33,7 +34,7 @@
     "tape": "^5.3.1"
   },
   "engines": {
-    "node": ">=10.12.0"
+    "node": ">=12.20.0"
   },
   "homepage": "https://standardjs.com",
   "keywords": [
--- a/test/api.js
+++ b/test/api.js
@@ -1,9 +1,9 @@
-const standard = require('../')
-const test = require('tape')
+import test from 'tape'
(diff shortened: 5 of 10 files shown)


bin/cmd.js
```diff
@@ -5,8 +5,12 @@ var match = process.version.match(/v(\d+)\.(\d+)/)
 var major = parseInt(match[1], 10)
 var minor = parseInt(match[2], 10)
 
-if (major >= 11 || (major === 10 && minor >= 12)) {
-  require('standard-engine').cli(require('../options'))
+if (major >= 12 || (major === 12 && minor >= 20)) {
+  import('standard-engine').then(mod => {
```
Divlo, line 9: This might be small, but to improve clarity, I would not call this `mod` but `module` or `standardEngine` instead.

What do you think ? :smile:

options.js
```diff
@@ -1,15 +1,21 @@
...
+import { dirname, join } from 'node:path'
+import { readFileSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
+import eslint from 'eslint'
 
-module.exports = {
-  bugs: pkg.bugs.url,
+const __dirname = dirname(fileURLToPath(import.meta.url))
+const json = readFileSync('./package.json', 'utf8')
+const pkg = JSON.parse(json)
```
Divlo, line 8: In my opinion, that would be clearer if we use the [URL API](https://nodejs.org/dist/latest-v16.x/docs/api/url.html) directly, instead of using this technique to have `__dirname`, that should be a workaround not a long term solution.

You could do it like so :
```js
import { URL } from 'node:url'
import fs from 'node:fs'

const pkgURL = new URL('./package.json', import meta url)
const pkgJSON = await fs.readFileSync(pkgURL, { encoding: 'utf-8' })
const pkg = JSON.Parse(pkgJSON)
```

**Note:** I don't think `pkg` is a great variable name (I know, it's the old name), we could probably name it `packageJSON` or something more clearer.

test/api.js
```diff
@@ -1,9 +1,9 @@
-const standard = require('../')
-const test = require('tape')
+import test from 'tape'
+import linter from '../index.js'
```
Divlo, line 2: Why are you renaming from `standard` to `linter` ?
I'm not against this change, just wanting to know if there's a specific reason.

test/cmd.js
```diff
@@ -1,8 +1,11 @@
-const path = require('path')
-const test = require('tape')
-const crossSpawn = require('cross-spawn')
+import { dirname, join } from 'node:path'
+import { fileURLToPath } from 'node:url'
```
Divlo, line 2: Same comment as above, we could use the [URL API](https://nodejs.org/dist/latest-v16.x/docs/api/url.html).

Divlo (changes requested):
Thank you for your contribution!
Indeed, this change will be needed somehow one day.

To migrate to ESM, we need to drop support for Node.js v10.
But that would require a **BREAKING CHANGE**, it's worth pointing out.
We already planning to drop support for Node.js v10 in `standard-engine`.
That should be a big problem.
But you forgot to remove Node.js v10 from the CI so the checks are failing.

Please see my comments.

bin/cmd.js
jimmywarting: sure, can change it. 
just started naming things mod just b/c Deno [did it](https://deno.land/manual/contributing/style_guide#do-not-use-the-filename-codeindextscodecodeindexjscode)... short for module

test/api.js
jimmywarting: I don't know. can revert.

test/cmd.js
jimmywarting: will fix things tomorrow

package.json
```diff
@@ -33,7 +34,7 @@
     "tape": "^5.3.1"
   },
   "engines": {
-    "node": ">=10.12.0"
+    "node": ">=12.20.0"
```
LinusU, line 37: Would prefer 👇 here, ref: https://github.com/sindresorhus/meta/discussions/15#discussioncomment-688195

```suggestion
    "node": "^12.20.0 || ^14.13.1 || >=16.0.0"
```

bin/cmd.js
```diff
@@ -5,8 +5,12 @@ var match = process.version.match(/v(\d+)\.(\d+)/)
...
+    import('../options.js').then(({ default: options }) => {
+      mod.cli(options)
+    })
+  })
```
LinusU, line 13: The purpose of this file is to run on really old versions of Node.js as well, that's why it's using e.g. `var` instead of `const`...

```suggestion
  import('standard-engine').then(function (engine) {
    import('../options.js').then(function (options) {
      engine.cli(options.default)
    })
  })
```

index.js
```diff
@@ -1,5 +1,7 @@
 /*! standard. MIT License. Feross Aboukhadijeh <https://feross.org/opensource> */
-const Linter = require('standard-engine').linter
-const opts = require('./options')
+import engine from 'standard-engine'
```
LinusU, line 2: Would be nice to convert `standard-engine` to ESM also, and then be able to use `import { Linter } from 'standard-engine'`

index.js
```diff
@@ -1,5 +1,7 @@
...
+import opts from './options.js'
```
LinusU, line 3: Not related to this PR, but I think it would be nice to rename `opts` to `options` ☺️

options.js
```diff
@@ -1,15 +1,21 @@
...
+const { bugs, homepage, version } = pkg
```
LinusU, line 9: I personally prefer these to be referred to as e.g. `pkg.verison` below, one less level of indirection to scan when reading the code...

test/external/clone.js
```diff
@@ -164,8 +168,8 @@ test('test github repos that use `standard`', t => {
   }), PARALLEL_LIMIT, err => {
     if (err) throw err
     if (argv.write) {
-      fs.writeFileSync(
-        path.join(__dirname, 'test.json'),
+      writeFileSync(
+        join(__dirname, 'test.json'),
```
LinusU, line 172: I think that all usages of `join` in this file could be replaced by using URL instead:

```suggestion
        new URL('test.json', import.meta.url),
```

tools/test-exists.js
```diff
@@ -2,9 +2,11 @@
  * Sanity check that all repos in test.json actually exist on GitHub
  */
 
-const get = require('simple-get')
-const testPkgs = require('../test/test.json')
-const series = require('run-series')
+import get from 'simple-get'
+import series from 'run-series'
+import { readFileSync } from 'fs'
```
LinusU, line 7:
```suggestion
import { readFileSync } from 'node:fs'
```

tools/test-exists.js
```diff
@@ -2,9 +2,11 @@
...
+
+const testPkgs = JSON.parse(readFileSync('../test/test.json', 'utf8'))
```
LinusU, line 9: The previous `require` call was relative to this file, but this is relative to the CWD (I think?)

Also, it seems like the file is in a folder called `external`? 🤔 

```suggestion
const testPkgs = JSON.parse(readFileSync(new URL('../test/external/test.json', import.meta.url), 'utf8'))
```

Tested this:

```sh
> new URL('../test/external/test.json', 'file:///home/user/standard/tools/test-exists.js')
URL {
  href: 'file:///home/user/standard/test/external/test.json',
  ...
}
```

test/external/clone.js
```diff
@@ -60,14 +64,14 @@ if (!argv.disabled) {
...
+  mkdirSync(TMP, { recursive: true })
 
   parallelLimit(pkgs.map(pkg => {
     const name = pkg.name
     const url = `${pkg.repo}.git`
-    const folder = path.join(TMP, name)
+    const folder = join(TMP, name)
     return cb => {
-      fs.access(path.join(TMP, name), fs.R_OK | fs.W_OK, err => {
+      access(join(TMP, name), R_OK | W_OK, err => {
```
LinusU, line 74: Not sure why this variable wasn't reused before 😄 

```suggestion
      access(folder, R_OK | W_OK, err => {
```

test/external/clone.js
```diff
@@ -89,7 +93,7 @@ test('test github repos that use `standard`', t => {
         }
 
         function gitClone (cb) {
-          const args = ['clone', '--depth', 1, url, path.join(TMP, name)]
+          const args = ['clone', '--depth', 1, url, join(TMP, name)]
```
LinusU, line 96: same here

```suggestion
          const args = ['clone', '--depth', 1, url, folder]
```

LinusU:
Could we change the Node versions in the `test-external` and `test-internal` workflow to the minimum supported versions?

```yaml
        node-version: [12.20.0, 14.13.1, 16.0.0]
```

I think that this is a better approach than testing on latest releases (e.g. `12.x`) since we could otherwise accidentally use something introduced in a later patch version which would then be a breaking change to our consumers...

bin/cmd.js
LinusU: Crap, `import` is a reserved word 😅 

Screenshot 2021-09-14 at 10 37 41

Use `eval`? 😄 

```suggestion
  eval('import("standard-engine")').then(function (engine) {
    eval('import("../options.js")').then(function (options) {
      engine.cli(options.default)
    })
  })
```

jimmywarting: Ha, eval/reserved word

> The purpose of this file is to run on really old versions of Node.js as well, that's why it's using e.g. var instead of const...

I figured it was

```js
eval('import("../options.js")')
```
won't this interfere with the cwd/path? feels like we would need a absolute path to options then...

index.js
jimmywarting: yea... but what should come first? the chicken or the egg? standard/standard-engine#251 (comment)

options.js
jimmywarting: I also don't like to destruct everything. I agree with you in some scenarios

index.js
LinusU: Yeah, I don't have an opinion on which comes first 👍

test/external/clone.js
jimmywarting: ty, this __dirname was the first solution that poped up in search results

index.js
Divlo: Indeed. :+1:
Actually, this is the reason for this PR (to have a smoother transition to ESM across the organization).
There is an open issue in `standard-engine`: https://github.com/standard/standard-engine/issues/251

jimmywarting: I like the naming convention doe. linter is a class so it should be exported/renamed to `Linter` from `linter`

package.json
jimmywarting: do we want to avoid the hole: "why isn't v15 supported"? question

index.js
Divlo: There is already a PR for that change in `standard-engine`: https://github.com/standard/standard-engine/pull/253

jimmywarting:
> Could we change the Node versions in the test-external and test-internal workflow to the minimum supported versions?

fixed

options.js
jimmywarting: I switched to use new URL
but i guess the name stays for now unless you have a better suggestion, `package` is a reserved word

```
const packageJSON = readFileSync(pkgURL, { encoding: 'utf-8' })
const package = JSON.parse(pkgJSON)
```

package.json
LinusU: Both Node.js 13 & 15 is end of life at this point, so I think it's an easy question to answer.

ref: https://nodejs.org/en/about/releases/

Are you suggesting that we stick to `>=12.20.0`? I think it would be good to indicate the minimum version of 14.x as well. Potentially we could do `^12.20.0 || >=14.13.1` but I think that it's better not to support 15.x since it's end-of-life...

jimmywarting:
ready for a 2nd wave of reviews now :)

options.js
```diff
@@ -1,7 +1,8 @@
-import { join } from 'node:path'
 import { readFileSync } from 'node:fs'
 import eslint from 'eslint'
 
+// eslintConfig.configFile have problem reading URLs and file:///
+const configFile = new URL('./eslintrc.json', import.meta.url).toString().slice(7)
```
LinusU, line 5: Can we use `fileURLToPath` (`import { fileURLToPath } from 'node:url'`) instead of `slice`?

```suggestion
const configFile = fileURLToPath(new URL('./eslintrc.json', import.meta.url))
```

test/external/clone.js
```diff
@@ -7,22 +7,23 @@
...
 import minimist from 'minimist'
 import parallelLimit from 'run-parallel-limit'
 import test from 'tape'
 
-const __dirname = dirname(fileURLToPath(import.meta.url))
-const json = readFileSync(join(__dirname, 'test.json'), 'utf8')
+const getAbsolutePath = (path, pwd = import.meta.url) =>
+  new URL(path, pwd).toString().slice(7)
+
+const testJsonPath = getAbsolutePath('test.json')
```
LinusU, line 19: (1/3)

I think it would be better to keep the variables as `URL` instead of strings as long as possible, instead of the other way around. That is, this variable can be an `URL` and be passed as that to `readFileSync` and `writeFileSync`.

```suggestion
const testJsonPath = new URL('test.json', import.meta.url)
```

test/external/clone.js
```diff
@@ -7,22 +7,23 @@
...
+const json = readFileSync(testJsonPath, 'utf8')
 const testPkgs = JSON.parse(json)
 
 const GIT = 'git'
 const NPM = 'npm'
-const STANDARD = join(__dirname, '..', '..', 'bin', 'cmd.js')
-const TMP = join(__dirname, '..', '..', 'tmp')
+const STANDARD = getAbsolutePath('../../bin/cmd.js')
+const TMP = getAbsolutePath('../../tmp/')
```
LinusU, line 26: (2/3)

```suggestion
const TMP = new URL('../../tmp/', import.meta.url)
```

test/external/clone.js
```diff
@@ -69,9 +70,9 @@ test('test github repos that use `standard`', t => {
   parallelLimit(pkgs.map(pkg => {
     const name = pkg.name
     const url = `${pkg.repo}.git`
-    const folder = join(TMP, name)
+    const folder = new URL(name, `file://${TMP}`)
```
LinusU, line 73: (3/3)

```suggestion
    const folder = new URL(name, TMP)
```

options.js
jimmywarting: yes

test/cmd.js
```diff
@@ -1,8 +1,7 @@
-const path = require('path')
-const test = require('tape')
-const crossSpawn = require('cross-spawn')
+import test from 'tape'
+import crossSpawn from 'cross-spawn'
 
-const CMD_PATH = path.join(__dirname, '..', 'bin', 'cmd.js')
+const CMD_PATH = new URL('../bin/cmd.js', import.meta.url).toString().slice(7)
```
LinusU, line 4: Same here

```suggestion
const CMD_PATH = fileURLToPath(new URL('../bin/cmd.js', import.meta.url))
```

test/external/clone.js
```diff
@@ -89,7 +94,7 @@ test('test github repos that use `standard`', t => {
         }
 
         function gitClone (cb) {
-          const args = ['clone', '--depth', 1, url, path.join(TMP, name)]
+          const args = ['clone', '--depth', 1, url, folder]
```
LinusU, line 97: Does this `URL` needs to be turned into a path?

```suggestion
          const args = ['clone', '--depth', 1, url, fileURLToPath(folder)]
```

jimmywarting: yea, folder where used on more places with child process, so i just casted the variable to path instead

Divlo (approved):
LGTM! :+1: 
Good job!

LinusU (approved):
Amazing 🤩

jimmywarting:
tried helping out in ts-standard also... but oh god what i hate typescript :P
sry, but someone else needs to take on that responsibility.

i support typing but only with proper jsdoc + vanila js

Divlo:
> tried helping out in ts-standard also... but oh god what i hate typescript :P
> sry, but someone else needs to take on that responsibility.
> 
> i support typing but only with proper jsdoc + vanila js
> i utilize typing with checkjs + allowjs but i don't use the typescript syntax that needs a compiler

@jimmywarting I updated https://github.com/standard/standard-engine/issues/251#issue-972660122 with the current status of the migration to ESM and I created an issue for each package that depends on `standard-engine`, when all engines like `ts-standard` etc will be migrated, we will be able to migrate `standard-engine` itself.

Concerning `ts-standard`, here's the issue: https://github.com/standard/ts-standard/issues/171
It is mainly blocked by `jest`.

PR #1744, opened 2021-11-15
Run external tests periodically

voxpelli:
To ensure that we catch errors in external modules early.

This runs the tests every third day at 4:47. The time is randomly chosen as is best practice for cron jobs.

GitHub Actions schedule syntax: https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#schedule

### DIFF
--- a/.github/workflows/test-external.yml
+++ b/.github/workflows/test-external.yml
@@ -5,6 +5,8 @@ on:
     branches: [master]
   pull_request:
     branches: [master]
+  schedule:
+    - cron: '47 4 1/3 * *'
 
 jobs:
   test-external:


voxpelli:
Visualization of the schedule: https://crontab.guru/#47_5_1/3_*_*

Divlo (commented):
We already have `dependabot` PRs quite often to catch `test-external` failures.
Even if it is not designed for that, it's already a kind of cron job, is it really necessary?

I'm not against the change, just questioning the usefulness of this.

LinusU (approved):

voxpelli:
> We already have `dependabot` PRs quite often to catch `test-external` failures. Even if it is not designed for that, it's already a kind of cron job, is it really necessary?

Yeah, currently PR:s like Dependabot becomes the first ones to catch these errors, making most of their failures false negatives.

Better to run dedicated tests for the external repositories, so we can more easily spot whether they are the cause or the PR:s are the cause.

PR #1745, opened 2021-11-15
Limit external runs to single Node version

voxpelli:
Picking `lts/*` to have it always run on the latest node.js LTS.

## Why?

Because it wastes less resources and linting should never differ between node.js versions for us, and if they are, then our internal tests will catch it.

### DIFF
--- a/.github/workflows/test-external.yml
+++ b/.github/workflows/test-external.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


Divlo (approved):
LGTM!

Indeed, as we run `test-internal` on all Node.js versions we support, it should already catch compatibility issues. :+1:

LinusU (approved):

PR #1747, opened 2021-11-19
Highlight errors rather than warnings on external CI

voxpelli:
A follow up to #1745 and #1744, purpose is to make it easier to spot which package it is that fails

### DIFF
--- a/.github/workflows/test-external.yml
+++ b/.github/workflows/test-external.yml
@@ -33,4 +33,4 @@ jobs:
         run: npm install
 
       - name: Run external tests
-        run: npm run test-external
+        run: npm run test-external-ci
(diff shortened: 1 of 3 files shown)


voxpelli:
For some reason no external tests are run here, while it did work locally, will have to revisit

voxpelli:
Related example: https://github.com/voxpelli/eslint-config/blob/main/.github/workflows/external.yml

voxpelli:
Seems like `tap-github-actions`  isn't working, should make replace with some custom little script that annotates instead. See eg: https://github.com/dependency-check-team/dependency-check/issues/202 and https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message

voxpelli:
I have no more time in the near time future to pour into this, closing to clean up

PR #1774, opened 2022-02-03
Improve pre-commit hook and docs

adamchainz:

[x] Documentation update

* Add the `description` field that will appear on https://pre-commit.com/hooks.html (I’m about to submit it).
* Describe what pre-commit does a little more.
* Describe the `pre-commit autoupdate` step required to pin to a specific version rather than `master`. Using `master` could be broken, and isn't reproducible between environments. Indeed pre-commit warns if you use `master`:

  ```
  [WARNING] The 'rev' field of repo 'https://github.com/standard/standard' appears to be a mutable reference (moving tag / branch).  Mutable references are never updated after first install and are not supported.  See https://pre-commit.com/#using-the-latest-version-for-a-repository for more details.  Hint: `pre-commit autoupdate` often fixes this.
  ```

No issue.

Nope.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,5 +1,6 @@
 - id: standard
   name: standard
+  description: JavaScript style guide, linter, and formatter
   entry: standard --fix
   language: node
   files: "\\.(\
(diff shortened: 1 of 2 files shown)


LinusU (approved):
Not too familiar with how `pre-commit` works, but LGTM 👍

Divlo (approved):
Thank you! :smile:

PR #1775, opened 2022-02-03
fix: update `eslint-config-standard-jsx`

### LINKED ISSUE #1548: Parsing error: Cannot use keyword 'await' outside an async function
**What version of this package are you using?**

14.3.4

**What operating system, Node.js, and npm version?**

Node: 14.13.0
npm: 6.14.8
OS: macOS Catalina 10.15.5

**What happened?**

Given code:

```js
import { fileURLToPath } from 'url'
import fs from 'fs'
const file = await fs.promise.readFile(fileURLToPath(import.meta.url))
console.log(file.toString())
```

Output is:

```
standard: Use JavaScript Standard Style (https://standardjs.com)
  example.mjs:3:14: Parsing error: Cannot use keyword 'await' outside an async function
```

**What did you expect to happen?**

`Top-level await` is a stage 3 proposal (see https://tc39.es/proposal-top-level-await/) but is enabled by default in from Node 14.8.0 (see https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V14.md#2020-08-11-version-1480-current-codebytere). 

This is going to mean that linting with apps written for Node 14+ will be problematic, as top-level await is bound to be used. 

Since Node 14 is becoming Active LTS in a few weeks I think TLA should pass linting

**Are you willing to submit a pull request to fix this bug?**

yes

voxpelli:
We need https://github.com/standard/eslint-config-standard-jsx/pull/44 to fix #1548

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

Ensuring the `ecmaVersion` set in `eslint-config-standard-jsx` reflects the one set in `eslint-config-standard`

Fixes #1548

Check that this actually fixes the problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   "dependencies": {
     "eslint": "^8.8.0",
     "eslint-config-standard": "17.0.0-1",
-    "eslint-config-standard-jsx": "11.0.0-0",
+    "eslint-config-standard-jsx": "11.0.0-1",
     "eslint-plugin-import": "^2.25.4",
     "eslint-plugin-n": "^14.0.0",
     "eslint-plugin-promise": "^6.0.0",
--- a/test/validate-config.js
+++ b/test/validate-config.js
@@ -1,0 +1,16 @@
+import test from 'tape'
+import standard from '../index.js'
+
+test('load config in eslint to validate all rule syntax is correct', async function (t) {
+  t.plan(1)
+  const code = 'const foo = 1\nconst bar = function () {}\nbar(foo)\n'
+  const [result] = await standard.lintText(code)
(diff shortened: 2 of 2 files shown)


Divlo (commented):
Maybe we can add a test case like we did inside `eslint-config-standard` to avoid regression.

voxpelli:
> Maybe we can add a test case like we did inside `eslint-config-standard` to avoid regression.

Good idea, added the very same, but adapted for the `standard` api. + I tested that it failed before the update of the jsx module and passed after the update 👍

Divlo (approved):

PR #1806, opened 2022-04-20
Increase `daysUntilClose` for stale bot

voxpelli:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain:

Make stale bot a bit less aggressive for now

Changed from 14 days between comment and close to 60 days

Gives us all more time to act on things

Just whether we want this change or not

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -4,7 +4,7 @@
 daysUntilStale: 180
 
 # Number of days of inactivity before a stale issue is closed
-daysUntilClose: 14
+daysUntilClose: 60
 
 # Issues with these labels will never be considered stale
 exemptLabels:


Divlo (commented):
Is an inactive issue no longer an issue?
It might be better to basically disable it completely and that we don't use a stale bot and let us (maintainers) manage issues.

I might be missing something, but is a stale bot really useful? What problem does it solve?
Are humans not capable to close issues when they are no longer relevant?
There are many repos on GitHub that use a stale bot but why, what does it bring?

I didn't have a opinion for this PR: https://github.com/standard/standard/pull/1734 but now I'm wondering if it's really useful.
Feel free to explain a little bit more why for this bot.
Thanks.

voxpelli:
I agree with you, I never use stale bots myself, I think an issue can be valid even if it's stale.

LinusU:
I personally feel very discouraged when my issues are closed by stale bots. I'm +1 on removing it

voxpelli:
@LinusU Seems like its a Probot app rather than a GH workflow, so I can't remove it as I'm not an admin, and we can't remove it through a PR.

Can you remove it and then close this?

Also: Maybe we could remove all Probot apps and instead use GH Actions?

LinusU:
Removing it seems to remove it from all 26 repositories within the Standard organization 😅 

But I guess that that is what we want, right?

voxpelli:
> Removing it seems to remove it from all 26 repositories within the Standard organization 😅
> 
> But I guess that that is what we want, right?

Yeah, I would say so

> > Also: Maybe we could remove all Probot apps and instead use GH Actions?
> 
> I think this sounds like a good idea 👍

There is already some precedent in that with the lock script: https://github.com/standard/standard/blob/master/.github/workflows/lock.yml

PR #1811, opened 2022-05-18
docs: add dotenv logo to README

motdotla:

[x] Documentation update
[ ] Bug fix
[ ] New feature
[ ] Other, please explain:

Added dotenv logo to the README

The dotenv library and its supporting libraries like dotenv-expand and dotenv-vault have been long time users of standard. We adopted standard in 2015. https://github.com/motdotla/dotenv/commit/9292ac143cc3429c6f6ecb883ac3a271c60613b7

Nope, that's it. Have a great day! 🌤


Divlo (approved):
Sure, thanks!
I think you can also add `dotenv` to `test/external/test.json`, this allows us to see if we do breakage. We disabled a lot of them recently because they were packages without maintenance over several years, so we might add new ones. :smile:

motdotla:
> Sure, thanks! I think you can also add `dotenv` to `test/external/test.json`, this allows us to see if we do breakage. We disabled a lot of them recently because they were packages without maintenance over several years, so we might add new ones. 😄

@Divlo looks like it is already there on line 21: https://github.com/standard/standard/blob/master/test/external/test.json#L21

The URL is the same and will always be the same.

PR #1815, opened 2022-05-25
docs(rules): fix markdown formatting issue

ddiu8081:

[x] Documentation update
[ ] Bug fix
[ ] New feature
[ ] Other, please explain:

In `no-multiple-empty-lines` part, I added an indent at the position of the blank line. This will fix the error as shown in the picture below on the left (a single block of code is incorrectly parsed into two parts, and the next section is incorrectly wrapped into the code block).

Screenshot

Link: https://standardjs.com/rules-zhcn.html and other languages.

Reason: This code block is located in an unordered list, so it should retain indentation, otherwise it would terminate the list.

Notes: `RULES.md` has already be fixed in https://github.com/standard/standard/commit/c4fa50760698e61b6aa24e988a4e3fbf001210ee, but I think adding indent is the less modified option.

### DIFF
--- a/docs/RULES-esla.md
+++ b/docs/RULES-esla.md
@@ -203,8 +203,8 @@ La mejor manera de aprender acerca de `standard` es instalarlo darle una prueba 
   ```js
   // ✗ evitar
   var value = 'hello world'
-
-
+  
+  
   console.log(value)
   ```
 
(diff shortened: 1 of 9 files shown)


LinusU (approved):
Thanks!

Issue #1814, opened 2022-05-24
v17 no longer respects `env` specified in `package.json`


GaryGSC:
A minimal reproduction is available at the repo [`garygsc/standardjs-regression`, which has two commits](https://github.com/GaryGSC/standardjs-regression/commits/main).

**What version of this package are you using?**
There's a regression when switching from `16.0.4` to `17.0.0` that wasn't documented in the changelog.

**What operating system, Node.js, and npm version?**
OS: `ubuntu-latest` (20.04.4)
Node.js version: `16.x` (v16.15.0, which comes with npm v8.5.5)

**What happened?**
Standard [exited with exit code 1 and the following errors](https://github.com/GaryGSC/standardjs-regression/runs/6580090459?check_suite_focus=true):
```
standard: Use JavaScript Standard Style (https://standardjs.com)
  /home/runner/work/standardjs-regression/standardjs-regression/index.test.js:1:1: 'describe' is not defined. (no-undef)
  /home/runner/work/standardjs-regression/standardjs-regression/index.test.js:2:3: 'test' is not defined. (no-undef)
  /home/runner/work/standardjs-regression/standardjs-regression/index.test.js:3:5: 'expect' is not defined. (no-undef)
```

**What did you expect to happen?**
Because the `package.json` [specified](https://github.com/GaryGSC/standardjs-regression/blob/main/package.json#L17-L21)
```json
"standard": {
  "env": {
    "jest": true
  }
}
```
we expected exit code 0 with no errors, like [what happens with `standard@16.0.4`](https://github.com/GaryGSC/standardjs-regression/runs/6580061786?check_suite_focus=true#step:5:7).

**Are you willing to submit a pull request to fix this bug?**
Yeah. It looks like [`standard-engine`](https://github.com/standard/standard-engine)'s the repo that [handles creating an ESLint config from the `package.json`](https://github.com/standard/standard-engine/blob/c3a7f17ec6a3ddee6ef64dbde5be69553d4a5c39/lib/resolve-eslint-config.js#L154), so this issue might be in the wrong place.

klaemo:
Looking at the [resolve-eslint-config.js](https://github.com/standard/standard-engine/blob/c3a7f17ec6a3ddee6ef64dbde5be69553d4a5c39/lib/resolve-eslint-config.js#L154) code you linked it seems like the config syntax changed slightly to:

```json
"standard": {
  "env": ["jest"]
}
```

This also fixed it for me locally.

[...]

PR #1816, opened 2022-05-30
docs: update environment configuration section

klaemo:

[x] Documentation update
[ ] Bug fix
[ ] New feature
[ ] Other, please explain:

I added an example illustrating the `standard.env` property in `package.json`.

Standard's environment configuration syntax in `package.json` is slightly different to what is documented for ESLint. This changed in version `17` of standard. Before, it accepted the ESLint syntax.

References: https://github.com/standard/standard/issues/1814

## Standard
```json
{
  "standard": {
    "env": ["mocha"]
  }
}
```

## ESLint
```json
{
  "standard": {
    "env": {
      "mocha": true
    }
  }
}
```

In all honesty, I think the underlying issue should probably be fixed instead 😅

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,6 +584,16 @@ Or, run:
 $ standard --env mocha
 ```
 
+Or, add this to `package.json`:
+
+```json
+{
+  "standard": {
+    "env": ["mocha"]
+  }
+}
+```
+
 Where `mocha` can be one of `jest`, `jasmine`, `qunit`, `phantomjs`, and so on. To see a
 full list, check ESLint's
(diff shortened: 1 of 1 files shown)


LinusU (approved):
Neat 👍

LinusU:
Hmm, actually, I didn't understand that current ESLint versions still uses the object syntax. I thought that the change in Standard was due to upgrading to a newer ESLint version.

I think it would be better if we make sure that the "proper" way works instead 🤔 

This probably needs to be done in standard-engine

PR #1822, opened 2022-06-22
Fixed comparison at `bin/cmd`

Rudxain:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

Changed a comparison within an expression to what it seemed correct. The expression after the `||` tested for equality, but the `>=` op already did that, so I changed it to allow `===` to be useful

None. I searched and didn't find anything similar

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -5,7 +5,7 @@ var match = process.version.match(/v(\d+)\.(\d+)/)
 var major = parseInt(match[1], 10)
 var minor = parseInt(match[2], 10)
 
-if (major >= 12 || (major === 12 && minor >= 20)) {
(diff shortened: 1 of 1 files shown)


voxpelli:
Thanks @Rudxain 🙏 

I will hold off on merging this as I just did a rewrite to make use of [`version-guard`](https://github.com/voxpelli/version-guard) and that way remove the possibility for these kinds of errors going forward: #1829 1829

It also helps checking that the version checked against is actually the intended version to check against, as it in this case actually should be `12.22`, not `12.20`.

This also relates to #1828

Rudxain:
Ok. So, should I close the PR? Is it useful to keep it open?

voxpelli:
Let’s leave it open for now

PR #1829, opened 2022-07-17
Fix test of silent exit on unsupported node.js versions

### LINKED ISSUE #1828: Old test is failing on `master`
The old test (GitHub Actions) on `master` is currently failing, it seems like a npm issue, not able to install dependencies.

I think it's worth removing this test as this doesn't really "test" the usage of StandardJS.
There was already a discussion about that in this PR: https://github.com/standard/standard/pull/1681

We can also remove the workaround with `eval` and `import` in `bin/cmd.js`, it makes the code ugly and I don't think it is worth it.
And instead to check the version with `process.version`, we can actually start using `engines` property in `package.json` (ref: https://github.com/standard/eslint-config-standard/issues/211).

Basically `bin/cmd.js` becomes:
```js
#!/usr/bin/env node

import * as standardEngine from 'standard-engine'

import options from '../options.js'

standardEngine.cli(options)
```

Thoughts?
Much cleaner, isn't it? Why adding code, we don't need and that actually break over time?

voxpelli:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

* Used `version-guard` for the silent exit (this helped identify a mismatch in the version set in `bin/cmd.js` and the one in `package.json`)
* Used `.cjs` file ending to avoid having to resort to `eval()`
* Gathered all non`-index.js` files (since there's now a `cli.js` file as well) in `lib/`
* Added Node.js 4 to `old-test.yml`, to get a a signal of how far back any possible breakage goes
* Stopped caching dependencies as we don't really want these very old npm versions to mess around with the cache that's shared by the rest
* Stopped installing dev dependencies in `old-test-yml` to avoid problems on npm install (this fixed the tests)

Closes #1828 


.github/workflows/old-test.yml
```diff
@@ -30,17 +30,9 @@ jobs:
...
-        uses: actions/cache@v3
-        with:
-          path: ~/.npm
-          key: ${{ runner.os }}-node-${{ hashFiles('**/package.json') }}
-          restore-keys: |
-            ${{ runner.os }}-node-
-
       - name: Install dependencies
-        run: npm install
+        run: npm install --ignore-scripts --production
```
Divlo, line 34: Why do we now need to have `--ignore-scripts` there?
Did you find what caused the issue on `master` branch?

Also, should we set `ignore-scripts=true` in `.npmrc` at the root (as you're now setting this option everywhere when running `npm install`)?

bin/cmd.cjs
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
+
+require('version-guard')('../lib/cli.js', 12, 22)
```
Divlo, line 3: Does that mean, that for Node.js version `< 12.22`, we fail silently?
What happens, if we run it, with for example Node.js v13, a Node.js version `> 12.22` but that we don't support as per `engines.node` property in `package.json`, ref: https://github.com/standard/standard/blob/8d832307d9f4e5875c7c32487e596d29db2cf10f/package.json#L37

Divlo (changes requested):
Thanks for investigating how to fix the old tests on `master`.
Some comments need more clarification of what's going on.

While I still think, we should not do some "version guard" things, I don't mind for the moment if we continue to fail silently for the older Node.js versions. About that, I will continue to discuss that in #1828, so this PR only talks about the implementation, not the possible solutions. :smile:

bin/cmd.cjs
voxpelli: That would go beyond fixing the/hardening the current guard so i would say its out of scope, especially considering the mixed stances on whether version guards at all are a good idea

.github/workflows/old-test.yml
voxpelli: I removed the `--ignore-scripts` now.

I added it as an hypothesis that it could be the cause. Then kept it due to best practice. But you're right, its out of scope.

bin/cmd.cjs
Divlo: Sounds good to me, considering the behavior is the same as before. :+1:

Divlo (approved):
LGTM!

PR #1854, opened 2022-10-07
Test on Node 18

voxpelli:
We should have these tests run on Node 18

### DIFF
--- a/.github/workflows/test-internal.yml
+++ b/.github/workflows/test-internal.yml
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.22, 12, 14.17, 14, 16.0.0, 18]
+        node-version: [12.22, 14.17, 16.0.0, 18.0.0, latest]
       fail-fast: false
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test-internal.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.22, 12, 14.17, 14, 16.0.0, 16]
+        node-version: [12.22, 12, 14.17, 14, 16.0.0, 18]
```
LinusU, line 18: I think that we should either keep the current system:

```suggestion
        node-version: [12.22, 12, 14.17, 14, 16.0.0, 16, 18.0.0, 18]
```

Or if we feel it's too much, something like:

```suggestion
        node-version: [12.22, 14.17, 16.0.0, 18.0.0, latest]
```

voxpelli: Good feedback, tweaked 👍

Divlo (approved):

PR #1908, opened 2023-05-29
Prepare release 17.1.0

voxpelli:


Divlo (approved):
Thanks! :+1:

The external tests failing is not related to us, we might need to disable it.

package.json
```diff
@@ -15,24 +15,24 @@
...
-    "eslint-config-standard": "17.0.0",
+    "eslint": "^8.41.0",
+    "eslint-config-standard": "17.1.0",
     "eslint-config-standard-jsx": "^11.0.0",
-    "eslint-plugin-import": "^2.26.0",
-    "eslint-plugin-n": "^15.5.1",
-    "eslint-plugin-promise": "^6.0.0",
-    "eslint-plugin-react": "^7.28.0",
+    "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-n": "^15.7.0",
```
LinusU, line 22: Should we bump/allow 16.x here?

```suggestion
    "eslint-plugin-n": "^15.7.0 || ^16.0.0",
```

LinusU (approved):
👍

package.json
voxpelli: @LinusU We can't without bumping the required Node to 16, and that would require us to release as Standard 18

Marcy336:
> _ No description provided. _

PR #1953, opened 2023-11-18
docs: first draft governance proposal

rhettjay:


wesleytodd:
I think this is a good start, we should really take time to consider the agreement model. Consensus seeking can be difficult, especially when it comes to the rules we choose which might be highly contentious. I am not saying I strongly oppose it, but I think it might cause problems to take this approach so we should be sure we understand the tradeoffs before adopting it. After we get a general direction set I will probably have more detailed feedback on the wording.

rhettjay:
Closing as it is stale. Suggestion to look at [neostandard](https://github.com/neostandard/neostandard).

Issue #1948, opened 2023-11-14
Maintenance & Governance of standard


mcollina:
After a [long twitter thread](https://x.com/jsumners79/status/1724164412599050587) and looking at the contribution graphs, it seems that to keep supporting Standard users, we need to come up with some new governance.

I maintain _hundreds_ of packages relying on standard, Including the whole Fastify organization (discussion link https://github.com/fastify/fastify/issues/5159). It seems _simpler_ to help maintaining this than moving all of those to a different formatter / linter.

@feross [said](https://x.com/feross/status/1724236724463608043?s=20) he would trust me to help with governance and kickstart the work again.

Should we schedule a quick call?

@standard/team @voxpelli @jsumners

mightyiam:
A mob consisting of @rostislav-simonik , @jay-bulk and I are trying to adopt this project. We are working on it four hours weekly.

I say trying because we haven't yet received npm publish permission for anything other than eslint-config-standard-with-typescript.

Would you be interested in hopping in during one of our sessions?

Schedule here:
https://mobusoperandi.com/mobs/standard.html

mcollina:
The mob programming model does not really work for the rest of us. I don't think I would be able to join your mob call anytime soon.

mightyiam:
Sorry—just to clarify—the rest of us whom, please?

jsumners:
> Should we schedule a quick call?

Sure. I'll do what I can.

mcollina:
> Sorry—just to clarify—the rest of us whom, please?

Currently, no one is doing releases for this project, and it has essentially stalled.
I need this linter to be maintained and moved forward as I have hundreds of packages depending on it, and I offered @feross my help to set up a governance model for this project. In that long twitter thread it appeared that quite a few people would be happy to keep moving this forward, and after @feross tweet I opened a discussion.

@voxpelli has been doing releases for the last 2 years, and this initiative has its blessing as well.

Ultimately, I would prefer to move this forward vs forking, but forking is 100% an option.

cc @wesleytodd

jay-bulk:
@mcollina I'm the newest "member" to join the mob group. I'd be happy to hop on a call and fill in the gaps if the rest of the team is unable.

mightyiam:
Sorry. Of course I'll be happy to join whatever call you're organizing. I'm also sorry that our mob did not receive @feross' trust by now and that this is happening instead. @feross did, however, let us know that he will look into our past contributions during December in order to assess his faith in us.

[...]

PR #1958, opened 2023-12-10
test: fix test to ensure --fix attempts

rhettjay:
I believe the tests were/are failing for the reason that --fix attempts were not actually being tried. This is because argv.fix was not set before the runStandard function being called. This should fix that.


jay-bulk:
I'd like to note, it's possible I misunderstood the intent of these tests. They have been running long enough that it feels like this type of logic should have been caught earlier, especially given the level of talent contributing to this repo. @voxpelli or @LinusU perhaps either of you would be willing to chime in on this matter. No pressure. Just trying to knock out some low hanging fruit while we wait on #1948.

rhettjay:
Stale and unlikely to be merged

PR #1959, opened 2023-12-15
docs: explicit call out for when process.cwd is set

### LINKED ISSUE #1956: `standard.lintFiles()` doesn't use `process.cwd()`
### Files

- `package.json`

  ```JSON
  {
    "name": "testcase",
    "version": "1.0.0",
    "type": "module",
    "dependencies": {
      "standard": "17.1.0"
    }
  }
  ```

- `index.js`

  ```JavaScript
  import fs from "node:fs/promises";
  import process from "node:process";
  import standard from "standard";

  console.log(await fs.readFile("foo/bar.js", "utf8"));
  console.log(await standard.lintFiles(["foo/bar.js"]));

  process.chdir("foo");
  console.log(await fs.readFile("bar.js", "utf8"));
  console.log(await standard.lintFiles(["bar.js"]));
  ```

- `foo/bar.js`

  ```JavaScript
  var baz = 'qux'
  ```

### Steps

1. `npm install`
2. `node index.js`

## What I expected to happen

```
var baz = 'qux'

[
  {
    filePath: '/home/regseb/testcase/foo/bar.js',
    messages: [ [Object], [Object] ],
    suppressedMessages: [],
    errorCount: 1,
    fatalErrorCount: 0,
    warningCount: 1,
    fixableErrorCount: 0,
    fixableWarningCount: 1,
    source: "var baz = 'qux'\n",
    usedDeprecatedRules: [Getter]
  }
]
var baz = 'qux'


[
  {
    filePath: '/home/regseb/testcase/foo/bar.js',
    messages: [ [Object], [Object] ],
    suppressedMessages: [],
    errorCount: 1,
    fatalErrorCount: 0,
    warningCount: 1,
    fixableErrorCount: 0,
    fixableWarningCount: 1,
    source: "var baz = 'qux'\n",
    usedDeprecatedRules: [Getter]
  }
]
```
## What seems to have happened

```
var baz = 'qux'

[
  {
    filePath: '/home/regseb/testcase/foo/bar.js',
    messages: [ [Object], [Object] ],
    suppressedMessages: [],
    errorCount: 1,
    fatalErrorCount: 0,
    warningCount: 1,
    fixableErrorCount: 0,
    fixableWarningCount: 1,
    source: "var baz = 'qux'\n",
    usedDeprecatedRules: [Getter]
  }
]
var baz = 'qux'

/home/regseb/testcase/node_modules/eslint/lib/cli-engine/file-enumerator.js:320
                    throw new NoFilesFoundError(
                          ^

NoFilesFoundError: No files matching 'bar.js' were found.
    at FileEnumerator.iterateFiles (/home/regseb/testcase/node_modules/eslint/lib/cli-engine/file-enumerator.js:320:27)
    at iterateFiles.next (<anonymous>)
    at CLIEngine.executeOnFiles (/home/regseb/testcase/node_modules/eslint/lib/cli-engine/cli-engine.js:797:48)
    at ESLint.lintFiles (/home/regseb/testcase/node_modules/eslint/lib/eslint/eslint.js:551:23)
    at StandardEngine.lintFiles (/home/regseb/testcase/node_modules/standard-engine/index.js:89:41)
    at file:///home/regseb/testcase/index.js:11:28 {
  messageTemplate: 'file-not-found',
  messageData: { pattern: 'bar.js', globDisabled: false }
}

Node.js v20.10.0
```

Standard doesn't use the current directory. It uses the current directory when importing `"standard"`:

- https://github.com/standard/standard-engine/blob/12c96e5/index.js#L105
- https://github.com/standard/standard-engine/blob/12c96e5/index.js#L36
- https://github.com/standard/standard/blob/d1cc665/index.js#L5

rhettjay:
Add clarity to docs to resolve #1956.


wesleytodd:
Glad to see this ended with a docs change. Not sure how much it matters, but even this wording is a bit incorrect in that it is "at the time the class is instantiated" right? For nearly all users I think that should be import time, but I almost feel like this could generate a few questions from folks if they are using the class and passing a `cwd`. Not sure I have a good suggestion seeing as even on the initial issue report my brain skipped right over the important detail of `chdir` in the middle of the program but just wanted to point it out.

My low effort idea is to phrase it as `current working directory (default: process.cwd() at class instantiation time)`.

wesleytodd (approved):

jay-bulk:
I went back and forth on this locally before pushing up. I ended with import because of the PR on [standard-engine](https://github.com/standard/standard-engine/pull/330). You're right though, it'll probably add to the confusion. I wonder if perhaps we opt for yours with a guided hint:
`current working directory (default: process.cwd() at class instantiation time -- see standard-engine docs`
Thoughts?

wesleytodd:
Honestly I am not very opinionated, just wanted to share what caught my eye reading it. I like the wording in this last more explicit one, but I agree that if end users wanting to do this understand and use "at import time" more then we should go with that. @regseb any thoughts?

regseb:
Standard only exports an object with the `lintText()` and `lintFiles()` functions. [_Is there a Node.js API?_](https://github.com/standard/standard/tree/5ff8441#is-there-a-nodejs-api)

This is an instance of the `StandardEngine` class, but the user doesn't know it. He may think that the Standard code is:

```javascript
export default {
    "lintText": function (text, opts) {
        // Do something.
        return result;
    },
    "lintFiles": function (files, opts) {
        // Do something.
        return result;
    },
};
```

I think that with the term "_at class instantiation time_", the user may be confused: "_which class?_"

I find the wording "_at import time_" correct.

rhettjay:
Given the issues with governance I don't believe it's likely this will be merged. Closing the PR. Check out [neo-standard](https://github.com/neostandard/neostandard) as an alternative.

PR #1961, opened 2023-12-24
feat: move standard to node native test runner

jay-bulk:
In relation to https://github.com/standard/standard-engine/issues/277 I wanted to test out the node native test runner on [standard](https://github.com/standard/standard). Open to thoughts/opinions. @voxpelli you opened that original issue, your opinion would be appreciated. I have been poking at testing it on [standard-engine](https://github.com/standard/standard-engine) as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,8 +71,8 @@
   },
   "scripts": {
     "test": "npm run test-internal && npm run test-external",
-    "test-internal": "./bin/cmd.cjs --verbose && tape test/*.js",
-    "test-external": "tape test/external/*.js",
+    "test-internal": "./bin/cmd.cjs --verbose && node --test test/*.js",
+    "test-external": "node --test test/external/*.js",
     "update-authors": "./tools/update-authors.sh && hallmark --fix AUTHORS.md"
   },
   "funding": [
--- a/test/api.js
+++ b/test/api.js
@@ -1,16 +1,16 @@
-import test from 'tape'
+import assert from 'node:assert'
+import test from 'node:test'
 import standard from '../index.js'
 
 test('api: lintFiles', async (t) => {
-  t.plan(2)
-  const [result] = await standard.lintFiles(['bin/cmd.cjs'])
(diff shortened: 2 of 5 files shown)


wesleytodd:
The projects I have use `node:test` with so far I ended up getting a bit frustrated with the `.only` behavior, but other than that it worked really well. I wonder if you have run into this? Not that that should block this, just wanted to mention it. 

I have been thinking that I might change my workflow to move to "one test per file" or something to help me get over that hump to adopt it. Not sure if there are any other approaches folks have found to make it easier to work on individual tests at a time.

Also, as an aside: I had a good conversation with Isaac and he ended up writing [a blog post about test styles](https://blog.izs.me/2023/09/software-testing-assertion-styles/) I found interesting.

jay-bulk:
I haven't run into any issues with `.only` yet but my experience thus far with the test runner has so far been without complaints.

I feel like I read that article when Isaac posted it in September. For that matter, I wish you and Isaac (and others) would take the time to share these types of thoughts more often. It's nice to get inside the minds of the legends of the community outside of twitter. I often find my assumptions challenged (which is good and also why I contribute to open source, i.e. become a better engineer).

wesleytodd:
> I haven't run into any issues with .only yet but my experience thus far with the test runner has so far been without complaints.

Awesome! I should probably loop back around on it soon then.

> I wish you and Isaac (and others) would take the time to share these types of thoughts more often.

I also wish this! I am no where near the level of folks like Isaac and Matteo (but am blushing being included even this little way). I think part of the problem is the way you make that kind of impact often requires all of your time and energy for many years. I know I would need to make pretty big life changes to be able to spend more time writing (although someday that is a goal of mine). Honestly I think OSS is the door opener, keep it up and don't let any imposter syndrome stop you from contributing. For me personally it has been a massive learning opportunity and challenge which I have found really rewarding (not to mention the career opportunity it can unlock).

PR #1962, opened 2024-01-22
Add SearchApi logo to users

SebastjanPrachovskij:
Added the SearchApi logo to the user's list!

We're using standardrb and standard ❤️

PR #1969, opened 2024-07-05
Search API 

stalker00321:
Reverts standard/standard#1962


SebastjanPrachovskij:
We're still using standardrb. 

I'm unsure why someone anonymous would like to remove SearchApi from the users list. :)

voxpelli:
Closing based on previous comment

PR #1981, opened 2025-02-22
Install `standard` as end users would in GitHub Actions

cclauss:


.github/workflows/user_test.yml
```diff
@@ -0,0 +1,32 @@
...
+          - "npm install standard"
+          - "yarn add standard"
+
+    steps:
+      - uses: actions/setup-node@v4
+        with:
+          node-version: latest
+      - run: |
+          ${{ matrix.command }}
+          standard --version || true  # Fails with: 'standard: command not found'
```
voxpelli: If you install it locally, then you need to invoke it as `npx standard` or `./node_modules/.bin/standard`

If you want it to be available directly then you need to install it globally instead: `npm install -g standard`

cclauss:
```suggestion
          node_modules/.bin/standard --version
```

voxpelli:
What would a test like this bring?

cclauss:
Closing.  I found the problem that I was having in an unrelated repo.  Sorry for the distraction.

PR #1996, opened 2026-07-13
Fix typo in README: 'thier' -> 'their'

codewithsyedabdullah:
Fixes a typo where 'thier' should be 'their'


michaelprimatarov1-droid (commented):
6

cclauss:
I would avoid putting the misspelling in the commit title because many repos publish changelogs with the title of landed pull requests, which perpetuates the misspelling in the docs.  This means that typo discovery tools such as [`codespell`](https://pypi.org/project/codespell) will need to have that typo added to their `--ignore-words-list` and will become blind to future regressions.

